PR #2, opened 2015-10-22
Dependencies

tomhukins:
To get GADS running I needed to install its dependencies.  The dependencies I have added to the Makefile.PL were all I needed to get the application running, but it's possible I have missed some.

I also made a couple of other cosmetic changes to the repository that I hope you find useful.


abeverley:
Thanks Tom, much appreciated.

PR #3, opened 2015-10-26
Add a new dependency

tomhukins:


abeverley:
Thanks Tom. Glad you're catching these things ;-)

Issue #8, opened 2016-07-22
Failure with latst D2/D2::P::DBIC


ribasushi:
( I **really** hope this is not another red herring this time, please double check my oneliner before doing anything else)

During last test passes I upgraded my deps, and it seems [this line is no longer valid](https://github.com/ctrlo/GADS/blob/master/lib/GADS.pm#L72):
```
~$ perl -e 'use Dancer2::Plugin::DBIC qw(schema resultset rset)'
Specific subroutines cannot be exported from plugin at -e line 1.
BEGIN failed--compilation aborted at -e line 1.
```

Here is the complete version list:
```
~$ perl -MModule::Versions::Report -e 'use Dancer2::Plugin::DBIC qw(schema resultset rset)'
Specific subroutines cannot be exported from plugin at -e line 1.
BEGIN failed--compilation aborted at -e line 1.


Perl v5.16.2 under linux 
 Modules in memory:
  Attribute;
  Attribute::Handlers v0.96;
  Attribute::Handlers::_TEST_;
  Attribute::Handlers::UNIVERSAL;
  attributes v0.19;
  B v1.35;
  B::AV;
  B::BINOP;
  B::BM;
  B::COP;
  B::CV;
  B::Deparse;
  B::FM;
  B::GV;
  B::HE;
  B::HV;
  B::IO;
  B::IV;
  B::LISTOP;
  B::LOGOP;
  B::LOOP;
  B::MAGIC;
  B::NULL;
  B::NV;
  B::OBJECT;
  B::OP;
  B::PADOP;
  B::PMOP;
  B::PV;
  B::PVIV;
  B::PVLV;
  B::PVMG;
  B::PVNV;
  B::PVOP;
  B::REGEXP;
  B::RHE;
  B::RV;
  B::Section;
  B::SPECIAL;
  B::SV;
  B::SVOP;
  B::UNOP;
  base v2.18;
  Carp v1.36;
  Carp::Heavy v1.36;
  Class;
  Class::Method;
  Class::Method::Modifiers v2.06;
  Class::MOP;
  Class::XSAccessor v1.19;
  Class::XSAccessor::Array;
  Class::XSAccessor::Heavy v1.19;
  Config;
  constant v1.27;
  CORE;
  CORE::GLOBAL;
  Cwd;
  Dancer2;
  Dancer2::Core v0.201000;
  Dancer2::Core::DSL v0.201000;
  Dancer2::Core::Hook v0.201000;
  Dancer2::Core::Response;
[...]

PR #9, opened 2016-07-25
Fix incomplete rsrc reregister

ribasushi:
Clean up the mess I made in https://github.com/ctrlo/GADS/pull/1/files

This is the core problem with the codebase. Given I produced the original screwup it is only fitting I fix it for good. The diag has now been extended to make this easier to spot/rectify in the future. See the commit msg for the full text.

Cheers!

P.S. This is unrelated to https://github.com/ctrlo/GADS/issues/8, which is still somewhat unresolved.

### DIFF
--- a/lib/GADS/DB.pm
+++ b/lib/GADS/DB.pm
@@ -73,6 +73,10 @@ sub add_column
             };
         }
     );
+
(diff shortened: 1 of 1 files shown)


abeverley:
Thanks @ribasushi  - I love the comments!

PR #10, opened 2016-07-25
Fix for recent perls - my( ..., my  ) is fatal

ribasushi:
I have no idea who/why made this fatal in recent perl versions, but there it is :/


abeverley:
Thanks @ribasushi

PR #12, opened 2017-07-17
Sass support

jkva:
- add node and npm support
- add sass
- move branding colours to separate sass import
- replace references to primary colour with sass variable
- write early .jshintrc, add jshint support

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ pm_to_blib
 *.swp
 config.yml
 _Inline/
+.sass-cache/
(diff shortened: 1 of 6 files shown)


jkva:
The reason jshint is not running yet is because our vendor JS sits in the same directory as our own JS. We should consider either using a more flexible taskrunner (Gulp) or moving our vendor JS into a separate `js/vendor/` folder.

abeverley:
Thanks Job

PR #13, opened 2017-07-25
Presentation role, Person Contacts, RAG column

jkva:
This adds the `Role::Presentation` layer, which allows clear separation between business logic and presentation data. It is to be rendered by a view.

Furthermore this deals with issues **392** and **395**.

### DIFF
--- a/lib/GADS/Datum/Person.pm
+++ b/lib/GADS/Datum/Person.pm
@@ -247,31 +247,6 @@ sub _set_text
     $self->text($value || "");
 }
 
-sub html
-{   my $self = shift;
-    my @details;
-    return unless $self->id;
-    if (my $e = $self->email)
-    {
-        $e = encode_entities $e;
-        push @details, qq(Email: <a href='mailto:$e'>$e</a>);
-    }
-    my $site = $self->column->layout->instance->site;
-    if (my $t = $self->freetext1)
-    {
-        $t = encode_entities $t;
-        push @details, $site->register_freetext1_name.": $t";
-    }
-    if (my $t = $self->freetext2)
-    {
-        $t = encode_entities $t;
-        push @details, $site->register_freetext2_name.": $t";
-    }
-    my $details = join '<br>', @details;
-    my $text = encode_entities $self->text;
-    return qq(<div>$text<button class="person contact-details trigger" aria-expanded="false"><span aria-hidden="true" class="glyphicon glyphicon-info-sign"></span>contact <span class="visually-hidden">details</span></button><div class="person contact-details expandable">$details</div></div>);
-}
-
 sub as_string
 {   my $self = shift;
     $self->text // "";
--- a/lib/GADS/Datum/Person.pm
+++ b/lib/GADS/Datum/Person.pm
@@ -27,6 +27,8 @@ use namespace::clean;
 
 extends 'GADS::Datum';
 
+with 'GADS::Role::Presentation::Datum::Person';
+
 sub set_value
 {   my ($self, $value, %options) = @_;
     ($value) = @$value if ref $value eq 'ARRAY';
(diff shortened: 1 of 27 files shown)


lib/GADS/Datum/Rag.pm
```diff
@@ -66,6 +68,19 @@ sub write_value
     $self->write_cache('ragval');
 }
 
+sub as_grade
+{
+    my $self = shift;
+    my %mapping = (
```
jkva: I'd rather define this mapping in the application config and set it as `state` var at init

lib/GADS/Datum/Person.pm
```diff
@@ -265,11 +267,9 @@ sub html
         $t = encode_entities $t;
         push @details, $site->register_freetext2_name.": $t";
     }
-    my $details = join '<br>', map {encode_entities $_} @details;
+    my $details = join '<br>', @details;
```
abeverley: Can this whole html() function be removed now? At this point, $details is part HTML and part not-HTML, so think it would be better removed altogether than accidentally used.

lib/GADS/Datum/Rag.pm
abeverley: As a general rule I'm removing stuff from the config file, as the new multi-tenanted set up means that more flexibility is needed. It could potentially go in the database, but I think just leave it here for now.

lib/GADS/Datum/Rag.pm
```diff
@@ -66,6 +68,19 @@ sub write_value
...
+        a_grey   => 'undefined',
+        b_red    => 'negative',
+        c_amber  => 'average',
+        d_green  => 'positive',
+        e_purple => 'unexpected'
```
abeverley: I think the Bootstrap definitions are better (warning, danger etc). For example, amber is normally used as some sort of warning.

lib/GADS/Datum/Person.pm
jkva: @abeverley I'd like it to be - but would prefer to factor out the `html`, `withlinks` and `form_html` routines in a separate PR, so this one doesn't become too large.

The coming PR would be using the presentation layer in other data templates as well, not just the data table, which currently isn't possible since they still rely on the old structure coming from the controller.

lib/GADS/Role/Presentation/Datum/Curval.pm
```diff
@@ -0,0 +1,33 @@
...
+    return \@links;
+}
+
+sub presentation {
+    my $self = shift;
+
+    return {
+        type  => $self->column->type,
+        text  => $self->text,
+        id    => $self->id,
```
abeverley: Is id needed? It will cause a panic when called on a multivalue column.

public/css/ctrlo-bootstrap.css
```diff
@@ -1234,3 +1234,45 @@ small.search_term {
...
+  border-left-color: #800000; }
+
+.rag.average {
+  border-left-color: #eda400; }
+
+.rag.positive {
+  border-left-color: #008400; }
+
+.rag.unexpected {
+  border-left-color: #891e88; }
```
abeverley: Do these colours have to be used for accessibility? It would be nice to use the old colours if possible, as these align with the rest of the site's colour scheme/branding.

lib/GADS/Role/Presentation/Datum/Curval.pm
jkva: It's used in rendering the `datum` in the template - so if this is a bug, it needs to be fixed.  Thanks!

lib/GADS/Datum/Rag.pm
jkva: That's why I brought it up during yesterday's call: I don't know the textual representations of RAG colours, so feel free to defined correct ones.

abeverley: Great. Let's go with undefined / danger / warning / success / unexpected

lib/GADS/Role/Presentation/Datum/Curval.pm
abeverley: I don't think it's used in there - the only id I can see is the id of the link, not the id of the datum.

This is the problem code: https://github.com/ctrlo/GADS/blob/8e1f4cac06404a23c4306e0a028fb11d2dab6633/lib/GADS/Datum/Curval.pm#L204

I want to remove the single "id" properties from datums as they are no longer applicable with the new multivalue feature. This is an interim fix.

abeverley:
Merged into the "menu" branch as a158228

PR #14, opened 2017-07-27
Table column controls

jkva:
This sorts:

## #396
Data Overview - Column sorting control should be button, not anchor.

## #404
Data Overview - column controls ("send e-mail" trigger, et al) should be buttons


abeverley:
Merged into the menu branch as ea0cfc6

PR #15, opened 2017-08-02
Disclosure widget for large text columns

jkva:
This deals with the fact that any table can have `n` columns that we have no control over. Any column can contain text depending on its type, and this text can be very large. It may also contain links.

I ended up going for a disclosure widget, in which the column header name is added to provide proper context. The column content is disclosed in familiar style, with a minimum width. If the width exceeds the viewport, it gets adjusted in its position.

This is still a hack to patch a deeper problem; having columns with lots of PRs, or columns with large amounts of text are a UX anti pattern. Ideally GADS administrators should be aware of how to effectively compose tables.

Any of the other potential solutions ("detail views", "card view", "list of cards", "rotate the table") come with other issues. The main obstacle remains the need for an entire flexible and undefined table configuration system, of which the results remain cognitively accessible.

In the end, reducing cognitive load will be the responsibility of the table designer. We can only do so much without having knowledge of the actual data and user goals.

### DIFF
--- a/views/data_table.tt
+++ b/views/data_table.tt
@@ -92,9 +92,7 @@
                 [% END %]
             </td>
             [% FOREACH column IN record.columns %]
-                [% column_index = loop.index %]
-                [% col_header = columns.${column_index} %]
-                <td class="[% column.type %]">[% render_datum(column, col_header.name) %]</td>
+                <td class="[% column.type %]">[% render_datum(column) %]</td>
             [% END %]
         </tr>
         [% END %]
--- a/views/data_table.tt
+++ b/views/data_table.tt
@@ -34,7 +34,7 @@
     [% END %]
 [% END %]
 <div class="hscroll">
-<table class="table table-striped" id="datatable">
+<table class="table table-striped" id="data-table">
     <thead>
         <tr>
             [% sort = sort.shift %]
@@ -92,7 +92,9 @@
                 [% END %]
             </td>
             [% FOREACH column IN record.columns %]
-                <td class="[% column.type %]">[% render_datum(column) %]</td>
+                [% column_index = loop.index %]
+                [% col_header = columns.${column_index} %]
+                <td class="[% column.type %]">[% render_datum(column, col_header.name) %]</td>
             [% END %]
         </tr>
         [% END %]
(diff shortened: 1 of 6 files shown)


abeverley:
Thanks Job, I like this, and sorry for the slow response. I'm assuming you have considered this, but is there any way of showing a preview of the beginning of the text for those long text cells that are hidden to begin with?

jkva:
Hi Andy, can you post a screenshot to illustrate?

abeverley:
As below. The first row does not have any text shown, just the Show button.

screenshot_2017-08-21_22-23-18

jkva:
It's not that simple, I'm afraid. If I generate a preview, I need to use a tokeniser. I can't just split on whitespace because that might cut a link in half. 

Even if that's not the case, if I use css clipping, I could be visually hiding a link. Which is bad usability and fundamentally violates WCAG. Plus I'd need to make sure that the overflow is keyboard-focusable so it can be scrolled... etc. Hence the huge wall of explanation behind this PR in the initial comment.

Robust UI and UX is impossible when structure and context of data is unknown. That's the main problem. That's why even this disclosure widget is still a hack.

A tokeniser to turn e.g [blah blah blah blah blah](url) into [blah...](url) is still possible, but that seems to be overkill. Plus it doesn't remove the other problem of possibly very narrow columns on smaller viewports, which is also bad UX. This PR is an intermediate hack, so we can move forward.

views/data_table.tt
```diff
@@ -92,7 +92,9 @@
                 [% END %]
             </td>
             [% FOREACH column IN record.columns %]
-                <td class="[% column.type %]">[% render_datum(column) %]</td>
+                [% column_index = loop.index %]
+                [% col_header = columns.${column_index} %]
+                <td class="[% column.type %]">[% render_datum(column, col_header.name) %]</td>
```
abeverley: I think it would be better to add the column name to the ```column``` hash in ```GADS::Role::Presentation::Record```, then you could just do

    render_datum(column, column.name)

which I'm happy to do on merge, unless I've missed something?

Or even just do ```render_datum(column)``` and pull it out of the hash in ```render_datum```?

jkva:
@abeverley Good suggestion indeed. Have made the column name part of presentation, and have sorted the popover wrapping.

PR #17, opened 2017-09-04
Global responsive styling

jkva:
This is a branch where I work on changes to the global system layout. Currently the grid layout isn't utilised properly, and the application uses fluid-width globally, which does not aid readability.

I've reworked the basic navigation, and have made full-width vs gutter-width configurable on a per-page basis. Thus, the data table uses all available screen width, while others might not.

This is not supposed to be merged, but rather a point to start discussing which changes to make and how.


jkva:
While I was working on navigation, one thing that struck me was that the various menus appear to sit in 'thin air', and don't clearly communicate what they are, except to screen-reader users.

I've added unobtrusive labels, and have moved the 'admin' navigation menu into the user menu, which means they'll have the options available if they have admin permissions.

One thing I've not given a place yet is the "Home" link. It has no clear obvious place - I'll think about that one. 

image

abeverley:
Thanks Job. Looks great in principle. Given that "instance" and "table" are the same thing (the latter should be used) we will want to reconsider the labelling of the top menu, and in fact might even want to move the table editor menu into it. I will need to discuss with Lindsay.

abeverley:
Also, I have been reminded that one of our outstanding items to consider is the addition of navigation breadcrumbs to each page. Is that something you have considered, and if not can you factor into your work please?

jkva:
@abeverley Yep, breadcrumbs are on my list. The nice thing about those it that it solves the "home" navigation link position as well.

abeverley:
Brilliant, thanks.

InSadly:
Hi jkva, abeverley,
For what it's worth - 
(a) User menu - isn't this more often called 'my account', 'my menu', 'my options' or something similar to distinguish it from the management of Users?
(b) 'Instance' is a new and confusing term, Table is better - there may be more appropriate still
(c) The purpose of the label "data" is not clear. I think the layout is absorbing too much of the screen 
(d) Similar thought to (c) why use extra rows for (currently) "Instance, User menu, Data" when these would fit on the same row and so would compress the total area of the header?
(e) Many customers will like a logo and branding, is the blue line at the top and 'HEADER' of value?
jkva - be aware, I'm late to the party. Don't prioritise my views or respond: just consider whether you have considered the point and disposed of it to your satisfaction. I'm on the phone or skype if you want to talk.

abeverley:
> (e) Many customers will like a logo and branding, is the blue line at the top and 'HEADER' of value?

@InSadly I can answer that one: HEADER is optional, and used to classify the data if the customer requires it. The blue line is part of the Ctrl O branding (albeit the colour needs updating) but we could change it for the customer if they wanted different branding.

jkva:
@InSadly Cheers! Locally I've been changing wording - you're very much correct on (a) and (b).

> (d) Similar thought to (c) why use extra rows for (currently) "Instance, User menu, Data" when these would fit on the same row and so would compress the total area of the header?

Certainly something I'd be willing to work on - I wasn't aware that it was open to redesign.

Will spend some time tomorrow on a clearer navigation and post updates. Would also love to sit and discuss what to do with 'Views' since the user testing showed a clear blocker there.

abeverley:
Thanks Job. Could I ask you to resolve the conflicts before I merge please? Sorry!

jkva:
@abeverley I'd rather not have this merged, sorry.

> This is not supposed to be merged, but rather a point to start discussing which changes to make and how.

I expect to get this merged when we're all happy with changes, and in the meantime we can discuss matters on this PR.

jkva:
@abeverley the latest changes add early BreadCrumb support:

image

jkva:
This was continued in #22, which has since been merged into main tree. Closing.

PR #22, opened 2017-10-09
Global responsive styling rebased

jkva:
This is a continuation of #17; I've rebased against master and merged later work. This includes menu reworking and accessible breadcrumbs.

The menu does not get hidden on mobile view; this is something I consider though. Would like to discuss how you're seeing that; I can make a visual mockup.


abeverley:
Thanks @jkva - now merged into menu branch

PR #25, opened 2017-11-17
Add autocur field to configuration options

jkva:


abeverley:
Thanks Job. I've discovered some bugs in my original implementation, so I've fixed these after the merge.

PR #26, opened 2017-11-17
Early rewrite of custom bootstrap selector dropdown

jkva:
@abeverley things this still needs:

* proper responsive styling (mainly for scrolling)
* the field rendering logic needs to have a way to either render an associative label (as it does by default) or render something else
* connected backend handling

Ideally I'd like to clean this template up into a block for each different field type, so we have more control over rendering logic and such. It's hard to debug otherwise.

### DIFF
--- a/views/edit.tt
+++ b/views/edit.tt
@@ -86,7 +86,7 @@
                 $(this).find('.modal-body').load(loadurl);
             });
 
-            $(document).on("mousedown", ".moreinfo", function(e) {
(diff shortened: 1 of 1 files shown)


abeverley:
Thanks Job, I'll send some comments separately.

jkva:
I recall this being dropped in favour of #29.

abeverley:
Correct, thanks.

PR #27, opened 2017-11-17
Early version of global search control

jkva:
This is the first version of the global quicksearch control.

As we discussed by phone, this needs to be selectively displayed, and be clearly indicated as "record search" as opposed to contextual (e.g. "users") search.


abeverley:
I made a mess when trying to merge this and fix conflicts (one of the bugs I fixed earlier). I've merged manually as 689517f.

PR #28, opened 2017-11-17
Various improvements towards accessibility

jkva:
Basically a bunch of "things that I wanted to fix" commits. Accessibility fixes and restructuring.
Also includes the table header changes that I mentioned in today's phone call.


abeverley:
Thanks, I'm happy with this, but it has conflicts from a previous merge. The problem is obvious, but I'm worried that if I do a manual fix that I'll remove some changes from the other merge. Can you rebase please?

abeverley:
Just checking where we are with this PR - is a merge still required, or has the code been incorporated elsewhere?

jkva:
@abeverley sounds like a fun merger I can sort tomorrow.

jkva:
@abeverley Can you review and merge if you're happy?

abeverley:
Thanks, looking good, just a few of minor problems:

- The column description in the pop-up breaks in the middle of words. E.g. "their" in the screenshot.
- The "email each person" should only appear for columns of type "person".
- The "email each person" button does not appear to work for a person column.
- The character to the left of the "email each person" does not display on my computer. Is this a local font issue?

These can all be seen in this screen-shot:
image

And sorry, more conflicts (I assume as a result of recent merges)!

jkva:
@abeverley Should be all good now 👍

abeverley:
Thanks Job, almost there! A couple more comments:

- Is it possible to wrap the column titles if they are long? The one I am looking at from a live system (10 words) takes up about half the width of my screen. I realise there may be other considerations to be had in this regard, but thought I'd ask.

- The description text has character breaks when wrapped:

screenshot at 2018-02-26 23 31 19

jkva:
image

abeverley:
Some example column names:

- Owner
- Other owner
- Availability
- Was the professional factor in place?
- DFLD Factor
- 2c. Reason for override
- Index Record Change(s)
- 3d. Reason for the change
- Do you want to change the override to one not recommended above?
- Last edited
- 2b. What part of this record do you consider to be most effective?

jkva:
@abeverley so this would become

image

jkva:
This looks like a good intermediate solution which works with the browser compatibility needs.

abeverley:
image

abeverley:
Thanks Job

PR #29, opened 2017-12-23
Accessible multiselect widget

jkva:
Hi Andy,

This is a _work in progress_ of an accessible multiselect widget. Can you check it out and see what you think?

I've also made sure the sticky footer on the edit form is visually standing out more.

Talk to you soon!

Job

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ config.yml
 _Inline/
 .sass-cache/
 node_modules/
+*.swo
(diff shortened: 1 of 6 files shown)


abeverley:
Thanks Job, looking really nice. I haven't tested it completely as it's a work in progress, but this is definitely looking like a good solution. Only comment is that it might be nice to have the list of options in its own scrollable box in case there are a lot of them.

Other 2 thoughts are: this will also need to be applied to the enum data type (I think you know this already), and it would also be nice to break it out into its own plugin that could be released to the community.

PR #30, opened 2018-01-30
Issue #468, #467: header content positioning

jkva:
This PR:

* improves quicksearch styling and adds it to the page header
* adds responsive styling to the page header
* clarifies front-end workflow in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,3 +101,12 @@ bin/generate.pl # Generate random data
 bin/onboard.pl --take-first-enum new.csv # Import random data
 ```
 
+## Front-end workflow
+
+### CSS
+CSS is written in SCSS, and compiled.
+
+The main SCSS files live in `scss/`. A changes monitor and compiler is used by running `npm run watch`.
+
+### Javascript
+A bootstrapping snippet lives in `public/js/linkspace.js`.
--- a/public/css/ctrlo-bootstrap.css
+++ b/public/css/ctrlo-bootstrap.css
@@ -228,11 +228,11 @@ nav .dropdown-toggle {
 .quicksearch input {
   width: 70%;
   display: inline-block;
-  height: 28px; }
+  height: auto; }
 
 .quicksearch button {
-  font-size: 1.05em;
-  margin-top: -0.2em; }
+  font-size: 1.4em;
+  margin-top: -.1em; }
 
 .breadcrumbs {
   padding: .5em; }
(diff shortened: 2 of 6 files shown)


abeverley:
Thanks, looks much better. Comments:

- Parameter search_text needs changing to search in the controller to make it work
- We need some way to make it obvious that a user is viewing filtered records when a search is live. This used to be done by replacing the view name header with "search results", but that has now gone with the new layout. Otherwise users forget that they are viewing a set of search results.
- I think there needs to be an easier way to clear the search

jkva:
@abeverley 

> Parameter search_text needs changing to search in the controller to make it work

Sorted this view-side. 

> We need some way to make it obvious that a user is viewing filtered records when a search is live. > This used to be done by replacing the view name header with "search results", but that has now
> gone with the new layout. Otherwise users forget that they are viewing a set of search results.

Added it back in, amended with results counter

> I think there needs to be an easier way to clear the search

Besides the current "Clear" button?

abeverley:
Thanks, that's good, but sorry, one more comment:

The height of the search box is slightly too small for the text on my screen (the "p" is chopped):
image

Maybe the font could be made smaller, and that would also allow more text on the screen (I feel the box is slightly too narrow at the moment for the text a user might be searching for).

jkva:
@abeverley I've made the input larger:

image

abeverley:
Thanks. Sorry to keep this open, but I now think it looks too big for the overall bar (it's taller than the menu to the left). Maybe remove the border around it? Also, the height of the button and the size of its text doesn't look quite right - is there any easy way of just making it fit the same height as the textbox?

abeverley:
Actually, I've just been using Stripe, and they don't have a button at all. Is there any mileage in something like they do, or does that fall foul of accessibility?
image

jkva:
> Actually, I've just been using Stripe, and they don't have a button at all.

It's more of an usability thing. Assuming Stripe's search has no autocomplete, they're assuming the user knows that "enter submits the query".

The input also lacks a label (using a placeholder instead) which is bad accessibility as well as a spec failure.

We have some issues to contend with when it comes to this PR, let's go over it in our morning call.

abeverley:
Okay - this is what it looks for me: 
image

jkva:
image

jkva:
image

jkva:
@abeverley I've repositioned and aligned things, care to have a look?

image

abeverley:
Brilliant, thanks Job. Only minor point is that the bottom of the text is being chopped slightly, but I'd rather keep it like that and have the improved styling.

PR #32, opened 2018-01-31
Issue 466  layout settings tabpanel

jkva:
This is a working tab panel layout for field editing. The controls footer has not been touched as I'd like to have that consistent with the _Edit Record_ functionality.

Mainly I'd like to hear thoughts on this. Personally I think it works very well, it's accessible, and can be operated by mouse as well as keyboard (arrow-left, arrow-right to cycle tabs).

### DIFF
--- a/public/css/ctrlo-bootstrap.css
+++ b/public/css/ctrlo-bootstrap.css
@@ -1475,11 +1475,14 @@ small.search_term {
   border: 0; }
 
 .tab-interface [role="tab"][aria-selected="true"] {
-  background-color: #fff; }
+  background-color: #fff;
+  font-weight: bold; }
 
 .tab-interface [role="tab"]:focus,
 .tab-interface [role="tab"]:hover {
-  background-color: #99ccff; }
+  font-weight: bold;
+  text-decoration: dashed underline black;
+  outline: 0; }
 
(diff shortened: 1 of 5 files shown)


abeverley:
Very nice, thanks, a great improvement. Comments:

- When the page is scrolled down and a tab is clicked, the page jumps back to the top
- I think "this field is optional" should move to "basic settings"
- I think field-specific options (e.g. Text field options) should be moved to advanced

jkva:
@abeverley 

> When the page is scrolled down and a tab is clicked, the page jumps back to the top

That happens when you're scrolled down. When you go from a taller-than-viewport tab to a fits-in-viewport-tab, the browser screen adjusts, and it looks like you jump up, because there's no vertical viewport scrolling room anymore.

I'm not sure how to solve this elegantly, not without making the scroll part of the tabpanels, which is worse for usability.

> I think "this field is optional" should move to "basic settings"

Done.

> I think field-specific options (e.g. Text field options) should be moved to advanced

Done.

abeverley:
One other minor issue - the "group" drop-down in the permissions tab is not styled as per other drop-downs in the application.

jkva:
@abeverley Sorted the hover/focus bold text on tabs, and took care of the permissions group `select`.

PR #33, opened 2018-02-01
Only show field save button when relevant

jkva:
This should sort Brass issue #463 - duplicate save buttons in layout.


abeverley:
Thanks. I think the individual field edit page and the index could be separate templates, but that's a job for another day :)

jkva:
Yep, agree with you there. Lots to do!

PR #34, opened 2018-03-06
(Rebased onto menu) multiselect widget with enum support

jkva:
Hi Andy,

This is about ready, the one thing I can't figure out is why the values submitted aren't coming through properly. Inspecting post requests to the server yield expected results, but there seems to be some strange versioning going on when editing records.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ config.yml
 _Inline/
 .sass-cache/
 node_modules/
+*.swo
--- a/public/css/ctrlo-bootstrap.css
+++ b/public/css/ctrlo-bootstrap.css
@@ -1777,3 +1777,12 @@ small.search_term {
 
 .select-widget .current [hidden] {
   display: none; }
+
+.edit-controls {
+  min-height: 4em;
+  padding-top: .5em;
+  border: 1px solid #c8c8c8;
+  box-shadow: .5em .5em .5em .5em #000; }
+
+#edit-form {
+  margin-bottom: 4em; }
--- a/public/css/ctrlo-bootstrap.css
+++ b/public/css/ctrlo-bootstrap.css
@@ -1373,44 +1373,8 @@ small.search_term {
 #reveal_password:checked ~ .reveal_area {
   display: block; }
 
-<<<<<<< HEAD
 /* disclosure widget */
 .expandable {
-=======
-.select-widget .form-control {
-  height: auto;
-  margin-bottom: .2em; }
-
-.select-widget .form-control > * {
-  width: 6em;
-  display: inline-block;
-  vertical-align: top; }
-
-.select-widget .label {
-  font-size: 1em;
-  font-weight: bold;
-  margin: 0;
-  padding: 0; }
-
-.select-widget-label {
-  margin-bottom: .3em; }
-
-.select-widget button {
-  border: 1px solid #336699;
-  border-radius: 0;
-  font-family: Helvetica, sans-serif;
-  vertical-align: top;
-  margin-bottom: .5em; }
-
-.select-widget .change {
-  margin-top: .3em; }
-
-.select-widget button:hover,
-.select-widget button:focus {
-  background-color: #eeefff; }
-
-.select-widget .chevron {
->>>>>>> Support single/multi value selection widget
   display: none; }
 
 .trigger[aria-expanded="true"] + .expandable,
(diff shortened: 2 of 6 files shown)


jkva:
Hi @abeverley, 

Could you go over this for me and see what else you want to tweak? Based on my own testing this works as expected but I want to run it by you.

abeverley:
Great, thanks, will look shortly.

abeverley:
Thanks, looking great, but quite a few comments (sorry!):

- The bold font of the label of non-optional fields seems to have disappeared.
- Can long lists be in a scrollable box please, rather than making the page really long?
- It would be nice if users could click anywhere on the widget to open it up, rather than having to click the little "change" button (I find it more intuitive to click where the value is, as that's how normal drop-downs work). It's nice to have the button, but it would be good to be able to click on the whole box as well.
- Single-select curval field types do not seem to work (the field does not display), but it works okay if it is turned into a "multiple select" field.
- For indented fields, the change button seems to have wrapped - see screenshot below
- The border around the drop-down for the curval type is very short and doesn't go around all the values (I can send a screenshot if need be)
- Is it possible to fix the height of the non-expanded field so that it's the same height as other fields (e.g. textboxes) to keep the page neat?

image

Thanks!

jkva:
@abeverley One of your requests was a consistent field height, but that's incompatible with longer option values, even for single-select.

image

I can give the fields a _minimum_ height, and expand from there.

abeverley:
> One of your requests was a consistent field height

I might not have been clear - if possible I'd like a consistent field height for the widget when it's *closed*, so it looks consistent with all the other fields on the page. Is that possible?

jkva:
@abeverley I understand what you mean - and in the example above the _selected value_ doesn't fit.
So currently what I've done is give it a _minimum_ consistent height, and it expands when needed.

Alternatively I use ellipsis overflow (as done previously) but then I need a way for the user to see the entirety of its answer, and I need to check whether there's actually any overflow (detection with Javascript) which makes it more complex.

I would otherwise argue the point of the edit form is to edit and observe data first, aesthetic uniformity second; minimum consistent height with expansion when needed seems the simplest compromise at this point.

abeverley:
Okay, I'm happy with a taller field when there is too much text - that makes sense. I think really I just want it to "fit in" with the other fields; as it was I think it was a different height even with just one line of text.

jkva:
Yeah, agreed 👍  thanks!

jkva:
@abeverley Proposal for full-clickable widget:

image

jkva:
@abeverley This works for now (sans proper backend handling which I'll get to). Only concern I have is that it's a pain to work with when there are lots of options (such as selecting a Court, or an Area). In native components one can use first-letter-quicksearch which we don't have available here.

Can you play with the current implementation and tell me what you think?

abeverley:
Brilliant, very nearly there! Just some (hopefully) minor things:

- The "details" button should not appear for type enum (there are no other details to show)
- For the curval, it's a bit annoying when the drop-down collapses after clicking on "show details". You might be half-way down a long list, want to check the details, and then select that option without having to find it again.

Final nice to have (don't worry if you don't have the time right now):

- For the single-value selects, it would be nice if the drop-down collapsed after selecting a value (like a normal drop-down)
- It would be nice if a value could be selected by clicking anywhere in the text's "row". I found myself clicking to the right of the text (like a normal drop-down) and it disappeared without being selected. As an alternative, maybe nothing could happen instead of collapsing.

Great work - really pleased.

jkva:
@abeverley 

> The "details" button should not appear for type enum (there are no other details to show)
Covered by 69bfd6b

> For the curval, it's a bit annoying when the drop-down collapses after clicking on "show details". You might be half-way down a long list, want to check the details, and then select that option without having to find it again.
Covered by e2bc1f5

> For the single-value selects, it would be nice if the drop-down collapsed after selecting a value (like a normal drop-down)
Covered by 79051b3

> It would be nice if a value could be selected by clicking anywhere in the text's "row". I found myself clicking to the right of the text (like a normal drop-down) and it disappeared without being selected. As an alternative, maybe nothing could happen instead of collapsing.

I'm not sure what you mean? The checkbox as well as its label form the entire click target. You want to widen the click target to the entire box around it? I can - just want to make sure I understand that right.

abeverley:
> I'm not sure what you mean? The checkbox as well as its label form the entire click target. You want to widen the click target to the entire box around it?

Yes, exactly that. Some options might be quite narrow (e.g. Male / Female) but the widget is much wider, so it would be nice to be able to click to the right of the text value.

jkva:
@abeverley Ok, only special case there is when there's a `details` button. So it's a bit more logic but should be doable.

abeverley:
> So it's a bit more logic but should be doable.

Awesome, thank you.

jkva:
@abeverley give this one a try. I also added some hover background to the answers.

abeverley:
Brilliant. Sorry, final thing, can we have a default mouse pointer over the text in the collapsed widget (instead of the text cursor)?

Other than that, are you happy for me to merge?

jkva:
@abeverley done. Feel free to merge once you've checked that values are persisted properly - I recall having enum or curval acting unexpected there and I'd rather be sure.

PR #35, opened 2018-03-15
Reintroduce layout tabpanel CSS

jkva:
This brings back the tabpanel layout as removed by a recent merger. It's also been moved to its proper sass import file.


abeverley:
Thanks @jkva

PR #36, opened 2018-05-14
Dropdown widget dependent field

jkva:
Hi Andy,

This handles dependencies on fields, with all the logic separated out into `linkspace.js`.
One case I'm not sure I'm handling correctly is the `tree` column type, can you explain what the special case is around this column type so I can account for it?

This PR also sorts the floating data table header issue.

### DIFF
--- a/views/edit.tt
+++ b/views/edit.tt
@@ -380,8 +380,10 @@
 
     [%# Skip field if it's an approval request but a value not needing approval %]
     [% NEXT IF page == "approval" AND NOT (record.fields.$colid.has_value)  %]
+
     [%# Skip if it's a child record and the value doesn't exist %]
     [% NEXT IF record.parent_id AND NOT (record.fields.$colid.has_value) AND NOT layout.user_can("create_child") %]
+
     [%# Skip if it's a linked record and the value is linked %]
     [% IF record.linked_id AND dispcol.link_parent %]
         <input type="hidden" name="[% field %]">
@@ -416,7 +418,10 @@
         [% readonly = "" %]
     [% END %]
 
-    <div class="form-group">
+    <div class="form-group" data-column-id="[% dispcol.id %]" data-column-type="[% dispcol.type %]"
+        [% dependency && 'data-has-dependency="' _ parent.id _ '"' %]
+        [% parent && dependency && 'data-dependency="' _ depends.regex_b64 _ '"' %]
+        >
         [% label = dispcol.name | html_entity %]
         [% IF dispcol.optional OR page == "bulk" %]
             [% label_style = 'style="font-weight:normal !important"' %]
@@ -495,7 +500,7 @@
 [% END %]
 
 [% BLOCK print_field %]
-    <div class="input_holder">
+    <div class="input_holder" >
         <div class="row">
             <div class="[% IF dispcol.multivalue AND dispcol.has_multivalue_plus %]col-md-10[% ELSE %]col-md-12[% END %]">
             [% IF dispcol.type == "enum" %]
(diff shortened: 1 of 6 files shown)


abeverley:
> One case I'm not sure I'm handling correctly is the tree column type, can you explain what the special case is around this column type so I can account for it?

There is nothing particularly special in it itself, it's just that it doesn't behave like a normal form field so I had to handle it differently (e.g. no jquery change() event). I think now that you've broken everything out nicely, that the tree handler can sit with the rest of the code with the multiselect widget, and not be treated in any special way (other than how the widget itself needs to be interfaced with).

It's also worth noting that a tree can have multiple values, so I loop through each one checking it in turn (see ```get_selected``` in tree.tt). I noticed that for the multiselect widget, that you joined the values together with a space. I realise this is similar to how I was doing it with the old widget, but I actually think it would be best to loop each value in turn, the same as the tree.

Finally, I'm not sure whether this PR is ready for testing yet, but I noticed that ```updateState``` is not being called when a single-value field is changed.

Thanks!

jkva:
> Finally, I'm not sure whether this PR is ready for testing yet, but I noticed that updateState is not being called when a single-value field is changed.

They have their own handlers to distinguish from the multi-option select.

> not be treated in any special way (other than how the widget itself needs to be interfaced with)

Ok, so I'd need to write the code for that.

> It's also worth noting that a tree can have multiple values, so I loop through each one checking it in turn (see get_selected in tree.tt). I noticed that for the multiselect widget, that you joined the values together with a space. I realise this is similar to how I was doing it with the old widget, but I actually think it would be best to loop each value in turn, the same as the tree.

Makes sense. I'll change it accordingly.

abeverley:
>> Finally, I'm not sure whether this PR is ready for testing yet, but I noticed that updateState is not being called when a single-value field is changed.
> 
> They have their own handlers to distinguish from the multi-option select.

Okay. I couldn't seem to get the single-value field to display/hide any dependent fields, and the above was as far as I got with debugging, but I'll try again once you've made the latest changes.

jkva:
https://github.com/ctrlo/GADS/blob/master/views/tree.tt#L68 
^ is this debugging?

https://github.com/ctrlo/GADS/blob/master/views/tree.tt#L77
^ this doesn't seem to follow array-of-objects syntax, and seems incorrect wrt jstree api docs. Am I misreading this?

abeverley:
> https://github.com/ctrlo/GADS/blob/master/views/tree.tt#L68
> ^ is this debugging?

Ooops, yes, thank you. Removed with da118d3

> https://github.com/ctrlo/GADS/blob/master/views/tree.tt#L77
> ^ this doesn't seem to follow array-of-objects syntax, and seems incorrect wrt jstree api docs. Am I misreading this?

You're right, but that code is actually no longer used, so I've just removed it (read-only tree values are now just displayed as text rather than in the widget as read-only).

Thanks!

jkva:
@abeverley how do tree widgets get their values? Are those pulled from an ajax endpoint by the tree JS? I can't find code in the existing templates that generates the tree statically, at least.

abeverley:
> how do tree widgets get their values? Are those pulled from an ajax endpoint by the tree JS?

Yes, it's an AJAX call: https://github.com/ctrlo/GADS/blob/b89df75/views/tree.tt#L59

Endpoint is here: https://github.com/ctrlo/GADS/blob/b89df75/lib/GADS.pm#L1479

jkva:
Tree fields seem to work properly, but I'm running into an issue when
adding a display condition dependency. Is this a regression problem? It
seems the [save] action is missing from display conditions.

On Sat, Jun 2, 2018, at 8:26 PM, Andy Beverley wrote:
>> how do tree widgets get their values? Are those pulled from an ajax
>> endpoint by the tree JS?> Yes, it's an AJAX call:
> https://github.com/ctrlo/GADS/blob/b89df75/views/tree.tt#L59> Endpoint is here:
> https://github.com/ctrlo/GADS/blob/b89df75/lib/GADS.pm#L1479> — You are receiving this because you authored the thread. Reply to
> this email directly, view it on GitHub[1], or mute the thread[2].

Links:

  1. https://github.com/ctrlo/GADS/pull/36#issuecomment-394106955

abeverley:
> Tree fields seem to work properly, but I'm running into an issue when adding a display condition dependency. Is this a regression problem? It seems the [save] action is missing from display conditions.

Yes, I did notice that - I'm not sure how long it's not been working as I didn't test it before. I just updated the database manually for my testing (set ```display_field``` and ```display_regex``` values in the layout table).

I noticed one other problem - I'll add a comment in-line to this PR.

views/edit.tt
```diff
@@ -416,7 +417,12 @@
         [% readonly = "" %]
     [% END %]
 
-    <div class="form-group">
+    <div class="form-group" data-column-id="[% dispcol.id %]" data-column-type="[% dispcol.type %]"
+        [% dependency && 'data-has-dependency="' _ parent.id _ '"' %]
+        [% (parent && dependency) && 'data-dependency="' _ depends.regex_b64 _ '"' %]
```
abeverley: -        [% (parent && dependency) && 'data-dependency="' _ depends.regex_b64 _ '"' %]
+        [% (parent && dependency) && 'data-dependency="' _ dependency.regex_b64 _ '"' %]

PR #42, opened 2018-07-12
Add Vagrant config

ronaldwind:


ronaldwind:
Hi @abeverley , this helps setup the application environment and we use it to run your app over here. You can review and merge it, but only if you want to of course.

abeverley:
Thanks!

PR #43, opened 2018-07-12
499 – Styling of multi-select widget

ronaldwind:


abeverley:
Brilliant, looking great. Just 2 queries before I merge:

1. Just checking that the npm package lock file should be included in this PR? (I'm happy if it is)

2. Slightly outside the scope of the original request (and I'm happy to treat separate if you prefer), but would it be possible to add something around the border of the drop-down to make it stand out a bit more once opened? For example, a normal HTML form drop-down includes some shading in my browser, which makes it a bit clearer once opened against everything else on the page. (I tried to take some screenshots but the standard drop-down disappears when I hit print screen). Let me know if that doesn't make sense.

ronaldwind:
Hi,

1. The package-lock.json is nice to have around so all developers explicitly use the same packages. It is however not related to this issue per se. We could create a new PR for it if you'd like.

2. We could add some shadow like:
dropdownshadow

ronaldwind:
(As you can see I already have no visible radio buttons, and have a 'blank' option. I'll create PRs for that shortly.)

ronaldwind:
I've added the shadow, and updated this PR.

abeverley:
Brilliant, thanks. I might add a little more to make it stand out further, but I can do that now I see how it's done :)

PR #44, opened 2018-07-12
500 Correctly save field display conditions

ronaldwind:


ronaldwind:
@abeverley No need to edit the controller, this fix should be sufficient.

abeverley:
Perfect, thanks.

PR #51, opened 2018-08-05
Allow select widgets to break out of its panel

ronaldwind:
The solution as included in the PR is quiet simple. It allow html elements to go outside the boundaries of a panel. But since a style to actually hide overflowing elements was in place, this change might break other things. It might have been there for a reason I don't know of. Maybe you have any idea? If not, just keep in mind that if you see something outside a panel that should've been cut off/hidden, this might be the reason :)


abeverley:
Ah, that was easy :)

Not quite sure why it was like that - I've been back through the code and it appeared a long time ago, so I'll merge and we'll see how we get on. Thanks!

PR #59, opened 2018-09-03
Feature ie8 compat

ronaldwind:


abeverley:
Thanks for this. I found that I also had to check whether console was defined, otherwise I got an undefined variable error in IE8: d583e2e

PR #73, opened 2018-11-18
Improve autocomplete of select widgets

ronaldwind:


abeverley:
Already merged as separate commit.

PR #77, opened 2019-01-07
Webdriver table management

tomhukins:
In order to get closer to testing views and filters, I have implemented table management tests.


abeverley:
Thanks Tom, looks good.

PR #84, opened 2019-03-12
Call more suitable DBIC methods

tomhukins:
When we only want one row, find() makes more sense than search().

### DIFF
--- a/lib/GADS/Graph.pm
+++ b/lib/GADS/Graph.pm
@@ -41,11 +41,11 @@ has _graph => (
         $self->id or return;
         $self->id =~ /^[0-9]+$/
             or error __x"Invalid graph ID {id}", id => $self->id;
-        my ($graph) = $self->schema->resultset('Graph')->search({
+        my $graph = $self->schema->resultset('Graph')->find({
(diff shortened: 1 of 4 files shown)


abeverley:
Thanks for this Tom, but I'm not sure that find() is suitable for all the cases here. From the [docs](https://metacpan.org/pod/DBIx::Class::ResultSet#find): "_If no such constraint is found, find currently defaults to a simple search->(\%column_values) which may or may not do what you expect. Note that this fallback behavior may be deprecated in further versions._"

I agree that find() should definitely be used for those in the PR that are a search on a unique constraint, but I'm not sure about the others based on the above doc extract?

tomhukins:
> I agree that find() should definitely be used for those in the PR that are a search on a unique constraint, but I'm not sure about the others based on the above doc extract?

Good point: they don't all use a primary key.  I was relying on test coverage catching problems, but I realise my development data set won't have enough variety to test this effectively.  I've reworked the patch to only call `find()` when using a primary key.

abeverley:
Thanks Tom, perfect.

PR #88, opened 2019-03-21
Further DBIC simplification

tomhukins:
Using find() instead of search()->next() in this well-tested code seems
like a worthwhile low risk improvement.


abeverley:
Thanks Tom, looks good.

PR #114, opened 2019-09-10
no underline in buttons with href

markov2:
Buttons without href are triggered by Javascript: their contained text is not underlined.  There is no reason to have a different visual effect on buttons which are link based.


abeverley:
Thanks @markov2 - fixed with 392bb231cd4bab5d4b0f25af3b4383692f42129d (styling changes need to be done in scss files)

PR #133, opened 2020-01-07
Fix the WebDriver tests

tomhukins:
5ffe635 added a new submit button, which meant the old test code didn't
know which one to click.  This change makes the button selector more
specific.


abeverley:
Ooops, sorry! Thanks for the quick PR.

PR #150, opened 2020-07-31
Update the CPAN dependencies

tomhukins:
Commit 6f8dd03bb02ef replaced PhantomJS with Chrome, causing test to
fail in a development environment with the old modules installed.

The failing tests, t/001_base.t and t/002_index_route.t, only run in the
presence of a config.yml file, so our current CI setup did not notice
this problem.

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -59,7 +59,7 @@ our %prereq_pm = (
     'Tie::Cache'                                => 0,
     'Tree::DAG_Node'                            => 0,
     'WWW::Form::UrlEncoded::XS'                 => 0,
-    'WWW::Mechanize::PhantomJS'                 => 0,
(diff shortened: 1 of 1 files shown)


abeverley:
Great, thanks Tom.

abeverley:
There might be others too? E.g. ``Log::Log4perl`` (wanted to avoid it but Mechanize seems to required it)

tomhukins:
> There might be others too? E.g. `Log::Log4perl` (wanted to avoid it but Mechanize seems to required it)

Good point.  The code now uses Log4perl, but there's no dependency stated for it.  I'll add it at some point unless you get there before me :smiley:

abeverley:
Just done :)

PR #183, opened 2022-03-07
805 20220203 print pdf stack

stolman-digitpaint:
Add timeline PDF printer based on javascript


lib/GADS.pm
```diff
@@ -1660,7 +1659,7 @@ post '/print' => require_login sub {
     );
 };
 
-prefix '/:layout_name' => sub {
+prefix '/:layout_name' => sub {data_calendar
```
abeverley, line 1663: Thanks for the PR @stolman-digitpaint - I'm just checking whether the above ```data_calendar``` is meant to be there? It doesn't look quite right.

stolman-digitpaint: it's not supposed to be there, no. I've removed this issue earlier, but there was a merge conflict when i pushed some finalizing local changes, and in the resulting merge they seem to have come back. i'll fix it shortly

stolman-digitpaint: issue is resolved

abeverley:
Thanks @stolman-digitpaint looking great. We'll do some final thorough testing but looks good on the face of it.

PR #185, opened 2022-04-19

stolman-digitpaint:
Add table permissions page
Add table edit page
Add topic overview page
Add topic add/edit/delete page
Add field overview page
Add field add/edit/delete page


lib/GADS/API.pm
```diff
@@ -692,7 +692,7 @@ sub _post_add_user_account
     my $update_user;
     if ($id)
     {
-        $update_user = schema->resultset('User')->active->find($id)
+        $update_user = schema->resultset('User')->find($id)
```
abeverley, line 695: I'm not sure about the above change - is there a reason for removing the search for only in active (non-deleted, accepted) users?

stolman-digitpaint: Hi Andy,

During testing, only searching in active users prevented account requests from being accepted. The API endpoint will only be used by the wizard for creating new users, or accepting account requests. Edit user is done via the GADS.pm endpoint. Thats why i had to remove the active check.

```
sub active
{   my ($self, %search) = @_;

    $self->search({
        account_request => 0,
        'me.deleted'    => undef,
        %search,
    });
}
```

PR #187, opened 2022-06-17

stolman-digitpaint:


stolman-digitpaint:
Added the following bugfixes to releasse:

- Add import page
- Add import data page
- Add import rows page
- Add timeline printer under the Downloads menu
- Add timeline options under the Action menu
- Add standardized data-attributes template
- Add missing data.tt modals

PR #193, opened 2022-07-22

stolman-digitpaint:


lib/GADS.pm
```diff
@@ -4356,6 +4356,7 @@ sub _process_edit
 
     if (my $delete_id = param 'delete')
     {
+        # say STDERR 'delete';
         if (process( sub { $record->delete_current }))
```
abeverley, line 4360: There are a few debug statements - should these be removed?

views/view.tt
```diff
@@ -374,6 +374,8 @@
                         
                         <div class="form-group ">
                           [%
+                            INCLUDE fields/hidden.tt id="refers_to_instance_id" value="10";
+
```
abeverley, line 378: I'm not quite following the need for this hidden field - is it needed and should it be hard-coded to a value of 10?

lib/GADS.pm
stolman-digitpaint: Yes, these should be removed. I will include them in the next update

views/view.tt
stolman-digitpaint: It was originally needed to display the QueryBuilder element. The javascript trigger on the field has now changed, so it is possible that it can be removed. I will need to test this first

abeverley:
Thanks for the comments @stolman-digitpaint - I'll merge once you've pushed the updates.

views/view.tt
stolman-digitpaint: From what i understand, in the JS layer, we currently only have one way of rendering the querybuilder (as a standardized component). It looks for this hidden field to render / process the element. I can change it, but it will take some time that i feel could be used better for other features / bugs. Would you like me to fix this first before we process the merge?

abeverley: Thanks Erik. I don't think there is any need to fix this now in entirety. I was just more concerned about the hard-coded value of "10". It looked like this might be a bug - can you confirm that this should indeed be a hard-coded constant value?

PR #194, opened 2022-08-19

stolman-digitpaint:


abeverley:
Thanks Erik. I'll follow up with the changes to ```GADS::Filter``` as discussed.

PR #202, opened 2022-11-01
Avoid Dancer2 warnings

tomhukins:
Calling header() instead of response_header() emits the warning:

  DEPRECATED: header keyword. Please use the 'response_header' keyword instead of 'header'


abeverley:
Thanks Tom, been meaning to do this for a while!

PR #204, opened 2022-11-03

stolman-digitpaint:


lib/GADS.pm
```diff
@@ -1553,6 +1555,19 @@ any ['get', 'post'] => '/user/:id' => require_any_role [qw/useradmin superadmin/
         $values{view_limits_with_blank} = $view_limits_with_blank;
         $editUser = \%values;
     }
+    elsif (my $delete_id = param('delete'))
+    {
+        return forwardHome(
+            { danger => "Cannot delete current logged-in User" } )
+            if logged_in_user->id eq $delete_id;
```
abeverley, line 1562: I'll merge this anyway, but just to note that numeric comparisons should be done with ```==``` instead of ```eq``` (which does a text comparison. Probably doesn't make any difference here, but worth bearing in mind.

abeverley: I've just pushed 65ca2ca4 which fixed this, although I noticed it's actually like that in the current version!

PR #229, opened 2023-04-27
Uiux

Oliver-ctrlo:
Set alert text and icons from white to black.
Set the band colours to match the website 
Set the alert backgrounds and borders to be 2 - 3 shades lighter than the brand colours  (rather than 3 - 4 shades darker)
Set the on hover colour to match brand primary


abeverley:
Thanks Oliver. Just before I merge this, can you update the black to be a variable rather than coding the value directly into the alerts file please?

PR #231, opened 2023-07-03
Update Person.pm

Oliver-ctrlo:
Added the 'deleted' attribute to the list of person field values (so it can be referenced by calculated fields).


lib/GADS/Datum/Person.pm
```diff
@@ -145,6 +146,8 @@ has value_hash => (
                 team_id       => $value->{team_id},
                 title         => $value->{title},
                 value         => $value->{value},
+                deleted       => $value->{deleted},
+
```
abeverley, line 150: Please remove extra carriage return!

lib/GADS/Datum/Person.pm
```diff
@@ -381,6 +393,8 @@ around 'clone' => sub {
         team         => $self->team,
         title        => $self->title,
         text         => $self->text,
+        deleted      => $self->deleted,
+
```
abeverley, line 397: Ditto

lib/GADS/Datum/Person.pm
Oliver-ctrlo: No problem, removed in latest commit.

abeverley:
Hi Oliver, just thinking about this more, I am wondering if the new deleted value needs to be converted to a standard date format for the calculated field value. E.g. if you take a look at this line: https://github.com/ctrlo/GADS/blob/11d5b8fb7522d196fcdf054f949a36bf6a430b66/lib/GADS/Datum/Date.pm#L206 you will see that it calls the ```date_for_code``` function to convert the date into a standard format which includes year, month, day etc. In order to do that you would also need to include this library: https://github.com/ctrlo/GADS/blob/11d5b8fb7522d196fcdf054f949a36bf6a430b66/lib/GADS/Datum/Date.pm#L30

Also to do that, you might also need to convert it from an epoch value to a DateTime object. If that is starting to look too complicated, then we can take a look at it when I next visit?

PR #233, opened 2023-09-19
Patch 1

Oliver-ctrlo:
Added keyboard functionality to the pop-over records, so rather than being just on-click pop-ups can be displayed using space or enter as-well.

Made the mandatory symbol (*) red so it stands out more.


abeverley:
Thanks Oliver, great work. Keyboard functionality is great, I am happy to merge. I'm not sure about the mandatory symbol in red (sorry, I was going to add a comment to the issue a while ago) - I'll discuss.

PR #235, opened 2023-09-25
Uiux

Oliver-ctrlo:
Added Keyboard functionality to the record-popup and the 2 different popover events that can be found within a record-popup table row.

Then by disabling defaults and adding Stop-propagation to the popover's event, it will stop any bubbling or keydown events interfering between the popovers and the record-popup they are nested in.


abeverley:
Thanks Oliver, this looks good.

PR #240, opened 2023-09-27
Error in datum RAG status fix

droberts-ctrlo:
An error in the file meant that the status wasn't displaying correctly when viewing the field data - this is now fixed


abeverley:
Thanks Dave. Can you also add a HTML filter please, just in case there are any HTML entities in the description? You can do that simply with ```[% description | html %]```

PR #241, opened 2023-09-28
Fix for child tables display error

droberts-ctrlo:
Issue 1351 (Brass) - not removing "no data present" on data added to child table (and adding it again after all child data is removed).


src/frontend/components/modal/modals/curval/lib/component.js
```diff
@@ -74,6 +74,12 @@ class CurvalModalComponent extends ModalComponent {
         hidden.closest(".table-curval-item").replaceWith(row_cells)
       } else {
         $(`#curval_list_${col_id}`).find("tbody").prepend(row_cells)
+        $(`#curval_list_${col_id}`).find("tbody").find(".dataTables_empty").each(function() {
+          let element = $(this)
+          if(!element.is(":hidden")) {
+            element.toggle();
+          }
```
abeverley, line 81: As discussed, please try replacing all this with `hide()`

src/frontend/components/button/lib/component.js
```diff
@@ -55,17 +55,25 @@ class ButtonComponent extends Component {
...
       if ($btn.closest('.table-curval-group').length) {
         if (confirm("Are you sure want to permanently remove this item?")) {
-          $btn.closest(".table-curval-item").remove()
+          const curvalItem=$btn.closest(".table-curval-item");
+          const parent = curvalItem.parent();
+          curvalItem.remove();
+          if(parent && parent.children().length==1) {
+            const child = parent.children('.odd').children('.dataTables_empty');
+            if(child.is(':hidden')) {
+              child.show();
```
abeverley, line 64: Consider simplifying with `parent.find('.dataTables_empty').show()`

PR #242, opened 2023-10-02
Modal dismiss on key

droberts-ctrlo:
This is a horrible fix, but it's the only way I could make it work!

Walks up the tree to see if the button is in a modal, if it is, it calls submit directly, if not, it doesn't.

If you use `$button.parents().find('.modal-body')` it will find the modal if the button is in a modal, but if it isn't, it enters a loop that crashes the browser while looking for the element.


src/frontend/components/button/lib/component.js
```diff
@@ -186,9 +186,11 @@ class ButtonComponent extends Component {
   }
 
   submitRecord(ev) {
-    const $button = $(ev.target).closest('button')
+    const $button = $(ev.target).closest('button');
     const $form = $button.closest("form")
     const $requiredHiddenRecordDependentFields = $form.find(".form-group[data-has-dependency='1'][style*='display: none'] *[aria-required]")
+    //This is awful, but if I use a find on all parents to find the .modal-body, the "normal" submit button crashes the browser with an (almost) infinite loop
+    const $parent = $button.parent().parent().parent().parent().parent().parent().parent().parent().parent().parent().parent();
```
abeverley, line 193: See if you can get this working with a `closest()`

abeverley:
Please also see if you can make this generic to also work on the main record.

PR #243, opened 2023-10-09
Markdown preview and help

droberts-ctrlo:
- Created a preview component for the editing of topics 
- Created a popover "cheat sheet" with help for markdown
- Created a template (.tt) with Markdown cheat-sheet
- Modified popover style to enable scrolling
- Made popover scrollbar more inline with branding (re: above)


abeverley:
Thanks Dave, this is great. Just a few minor comments (these look like a lot, but just lots of minor issues!):

- Can you change "Preview" to "Description preview"?
- Can you make the preview code run on initial page render? So that when there is already a markdown description, it shows the preview
- I'm wondering whether there's a reason for using ```overflow-y: scroll``` instead of ```auto```?
- I think it might be useful to also include the standardised scrollbar styling that Firefox supports (see comments in https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar)
- I think for the preview box, that rather than using custom styling, it would be better to use a normal ```textarea``` component and make it readonly/disabled
- It looks like ```views/snippets/markdown.tt``` is not used? It looks like it would indeed be useful to include the help text in a snippet though, as having the information in ```popover_body``` directly makes the template difficult to read.
- Can I ask what the reason is for adding the new ```row``` and ```col``` divs please?

Thanks!

droberts-ctrlo:
Going to "respond" in order of comments :smile:
* Done
* Admittedly, it was a while ago, but when I was at uni, there were issues with Auto in some browsers, and I prefer to be explicit
* Done
* Is it possible to render HTML in a TextArea? If so, I'll change it, but I don't think it is :question:
* Initially, if I'm honest, I was uncertain how to do this - it is, however, now implemented
* The initial "layout" had a lot of extra rows and cols, which didn't appear to affect rendering, I managed to squash them down to what is included now with no impact on rendering

abeverley:
> * Admittedly, it was a while ago, but when I was at uni, there were issues with Auto in some browsers, and I prefer to be explicit

I think let's try the auto for now, and change it if it becomes a problem (just so the scrollbar doesn't appear unnecessarily in other pop-ups)

> * Is it possible to render HTML in a TextArea? If so, I'll change it, but I don't think it is ❓

Good point, probably not. Can you instead try form input classes? E.g. ```<div class="form-group"><div class="textarea form-control">...</div></div>

Thanks!

droberts-ctrlo:
> I think let's try the auto for now, and change it if it becomes a problem (just so the scrollbar doesn't appear unnecessarily in other pop-ups)
Already done :smile: 
> Good point, probably not. Can you instead try form input classes?
I tried this, and it appears the overflow and height of the form controls appears fixed (and makes a mess of all of it) - I have taken what I feel is "appropriate" from bootstrap's own files, and set them as the style for the preview.

PR #244, opened 2023-10-09
Changed "loading" card

droberts-ctrlo:
Loading card should now display on screen, rather than the centre of the table when searching


abeverley:
Thanks Dave. I think it would be better to avoid the use of ```!important``` if possible. To do this, you should be able to match the class exactly (```div.dataTables_wrapper div.dataTables_processing```) and then I think ```external.css``` and ```general.css``` will need swapping round in ```views/layouts/main.tt``` (it feels this would be the better way round anyway). Can you try that and check that it doesn't have any other impacts please?

Also, you should then be able to remove the other CSS properties, just leaving the new ```top``` parameter.

Finally, it doesn't really matter for now, but things like the update to ```.gitignore``` should probably be in a different PR.

Thanks!

PR #245, opened 2023-10-12
Button code refactoring

droberts-ctrlo:
I have created classes for each of the buttons as there was no commonality between them, and as such, the large class containing all the button code was unwieldy and difficult to maintain.

I am uncertain as to where the conflict has arisen from, apologies if this makes things difficult.


abeverley:
Thanks for this Dave, looks great. The PR seems to be have conflicts now though, presumably because of changes since you submitted it (sorry). Could you resolve the conflicts please so that I can merge?

abeverley:
Thanks Dave, but it seems to have some unwanted changes now (removal of Markdown and some extraneous changes to files such as general.css).

droberts-ctrlo:
Because of my merge from UIUX to this branch, I had to manually merge the Markdown code - this is the same as in the UIUX branch, but shows as a change here due to my branch not having these files originally (this is the merge-conflict I had, the markdown code wasn't in my version when I was working on buttons initially).

It's showing the files and information as added because it's new to my branch, but it remains in the UIUX branch (I think that makes sense)

abeverley:
Thanks Dave. I think the problem is that there are a couple of files that are missing the Markdown aspects. If you take a look at the files tab you'll see what I mean: https://github.com/ctrlo/GADS/pull/245/files

I'm happy to merge this manually if it's easier.

droberts-ctrlo:
Hi Andy, the entire `./src/frontend/components/button` directory is no longer present, I have moved all the buttons to `./src/frontend/components/buttons` (as there is more than one) - the package.json wasn't committed a moment ago (because out of habit I didn't commit it, even though it has the dependency to `marked` which is used for the rendering) - from what I can see, apart from that, the merge *appears* to me that there's nothing missing, but I'm open to the fact that I'm talking scribble (rubbish).

abeverley:
Thanks Dave. Sorry, I didn't mean the files are missing, just the Markdown aspects in the files. Anyway, looks good now, thanks. I'd normally keep the built JS/CSS files out of the PR (and avoid other changes such as changing package version numbers, unless directly related), but I'll merge this so that it's done. Many thanks.

PR #246, opened 2023-10-13
Asterisks now present

droberts-ctrlo:
Asterisks and styling now present on required multi-value fields and trees


abeverley:
Thanks Dave, this great. Just some pernickety comments:
- Can you indent the lines within the `IF` and `UNLESS` statements please?
- Personally I would use IF/THEN rather than UNLESS/THEN (and swap the statements) as I find it to be more readable. That is just personal preference though
- The Github workflow would be better in a separate PR, but happy to include it in this for the time being.

abeverley:
Perfect, thanks.

PR #248, opened 2023-10-17
Edit user updates

droberts-ctrlo:
- Added notes field to user on edit as well as open
- Allowed for rename of notes field
- Allowed for modification of placeholder text in notes field
- Allowed for modification of placeholder text in freetext fleids
- Modified notes field to allow to be textarea
- Edited template to allow for textarea data type


abeverley:
Thanks Dave, this is great. I realised I forgot a couple of things that are needed when doing a schema upgrade:
- Please copy all the fixtures from the previous version to the new version (e.g. from `share/fixtures/99` to `share/fixtures/100`
- The default SQL for creating the database needs to be updated (in `share/migrations/PostgreSQL/deploy/xx/001-auto.sql`). Take a look in a previous one, and you will see a number of fields that are type `citext`. These need to be changed in the new SQL that is automatically generated.
- In fact, I have just realised you need to also add all the above files to the PR that have been created in the `share` directory
- Finally, it would be great if the indenting of the `push @fields` statement can match the ones above ;-)

droberts-ctrlo:
I have copied all files as per the first point - there were a number of files that would have been overwritten, I have left these as they are. All other points have been actioned.

PR #251, opened 2023-10-18
Button revert

droberts-ctrlo:
I've tested it and deleted a user on my end, and it worked (my bad).


abeverley:
Thanks Dave. We can put the changes back in slower time.

PR #257, opened 2023-10-24
Unit test fixes

droberts-ctrlo:
Have made unit testing work - "test" folder contains stubs or full implementations for use within testing (as babel is used and doesn't "like" certain paths because of exclusion rules, causing tests to fail) - will explain if required in face-to-face meeting this week with AB


droberts-ctrlo:
Ignore, I merged my code in the wrong direction!

PR #261, opened 2023-10-31
Reporting complete

droberts-ctrlo:


public/css/external.css
droberts-ctrlo, line 1: Don't know why these are here - can't do a lot about it now, maybe cherry-pick?

public/css/general.css
droberts-ctrlo, line 1: This should be the same as the original

src/frontend/js/site.js
```diff
@@ -36,8 +36,8 @@ import TippyComponent from 'components/timeline/tippy'
 import TreeComponent from 'components/form-group/tree'
 import UserModalComponent from 'components/modal/modals/user'
 import ValueLookupComponent from 'components/form-group/value-lookup'
-import MarkdownComponent from "../components/markdown"
-import ButtonComponent from "../components/button"
+import MarkdownComponent from "components/markdown"
```
droberts-ctrlo, line 39: Just a typo, but I thought I'd fix it while I was here!

src/frontend/test/validation.js
```diff
@@ -293,7 +293,8 @@ const expandCardValidate = (field) => {
       field.scrollIntoView()
       $(this).off('shown.bs.collapse.foobar');
     })
-    $collapse.collapse('show')
+    if($collapse && $collapse.collapse)
```
droberts-ctrlo, line 296: Testing broke so I had to add this - it is only in the "test" stub - the original remains unchanged

.github/workflows/tests.yml
```diff
@@ -95,3 +95,21 @@ jobs:
         # TODO: Delete the line below and uncomment the line beneath that
         run: prove -lmrsv webdriver/t/login.t
         #run: prove -lmrsv webdriver/t
+
```
droberts-ctrlo, line 98: I know this would normally be in another pull, but the tests have been added to this commit, so I didn't see the point in modifying the file separately when it wasn't needed.

public/css/external.css
droberts-ctrlo: Fixed

public/css/general.css
droberts-ctrlo: Fixed

lib/GADS.pm
```diff
@@ -2753,6 +2755,290 @@ prefix '/:layout_name' => sub {
...
+            template 'report' => $params;
+        };
+
+        #add a report
+        any [ 'get', 'post' ] => '/add' => require_login sub {
+            if ( body_parameters && body_parameters->{submit} ) {
+                my $layout             = var('layout') or pass;
+                my $report_description = body_parameters->{report_description};
+                my $report_name        = body_parameters->{report_name};
+                my $checkbox_fields_full = body_parameters->{checkbox_fields};
```
abeverley, line 2817: Should be able to use: `my @checkbox_fields = body_parameters->get_all('checkbox_fields')` instead to get all values.

In general, it's better to use `body_parameters->get('foo')` instead of `body_parameters->{foo}`

lib/GADS.pm
```diff
@@ -90,6 +91,7 @@ use Dancer2::Plugin::Auth::Extensible::Provider::DBIC 0.623;
 use Dancer2::Plugin::LogReport 'linkspace';
 
 use GADS::API; # API routes
+use List::MoreUtils qw(uniq);
```
abeverley, line 94: Should not need any more, see below

lib/GADS.pm
```diff
@@ -1857,6 +1858,7 @@ any qr{/(record|history|purge|purgehistory)/([0-9]+)} => require_login sub {
     my ($return, $options, $is_raw) = _process_edit($id, $record);
     return $return if $is_raw;
     $return->{is_history} = $action eq 'history';
+    $return->{reports} = schema->resultset('Report')->load_all_reports($record->layout->instance_id);
```
abeverley, line 1861: Might be worth considering a new method to only return the reports, rather than the reports with the fields. You could even create a new method in `GADS::Layout` such as `$layout->reports`

lib/GADS.pm
```diff
@@ -2753,6 +2755,290 @@ prefix '/:layout_name' => sub {
...
     };
 
+    prefix '/report' => sub {
+
+        #view all reports for this instance
+        get '' => require_login sub {
+            my $user   = logged_in_user;
+            my $layout = var('layout') or pass;
+
+            if ( app->has_hook('plugin.linkspace.data_before_request') ) {
```
abeverley, line 2765: Is this hook needed?

lib/GADS.pm
```diff
@@ -2753,6 +2755,290 @@ prefix '/:layout_name' => sub {
...
+
+            my $fields = [
+                map $_->presentation(
+                    group            => $records->is_group,
+                    group_col_ids    => $records->group_col_ids,
+                    sort             => $records->sort_first,
+                    query_parameters => query_parameters,
+                ),
+                uniq @columns
+            ];
```
abeverley, line 2869: Replace with `$layout->all(user_can_read => 1)`

lib/GADS.pm
```diff
@@ -2753,6 +2755,290 @@ prefix '/:layout_name' => sub {
...
+
+            my $alert = GADS::Alert->new(
+                user   => $user,
+                layout => $layout,
+                schema => schema,
+            );
```
abeverley, line 2875: I don't think alerts are needed?

lib/GADS.pm
```diff
@@ -2753,6 +2755,290 @@ prefix '/:layout_name' => sub {
...
+
+            if ( body_parameters && body_parameters->{submit} ) {
+                my $layout               = var('layout') or pass;
+                my $report_description   = body_parameters->{report_description};
+                my $report_name          = body_parameters->{report_name};
+                my $checkbox_fields_full = body_parameters->{checkbox_fields};
+                my $instance             = $layout->{instance_id}
+                  if $layout && $layout->{instance_id};
+
+                error "NO INSTANCE" if !$instance;
```
abeverley, line 2910: Can remove - this will never happen

lib/GADS.pm
```diff
@@ -2753,6 +2755,290 @@ prefix '/:layout_name' => sub {
...
+
+        #Edit a report (by :id)
+        any [ 'get', 'post' ] => '/edit:id' => require_login sub {
+
+            if ( body_parameters && body_parameters->{submit} ) {
+                my $layout               = var('layout') or pass;
+                my $report_description   = body_parameters->{report_description};
+                my $report_name          = body_parameters->{report_name};
+                my $checkbox_fields_full = body_parameters->{checkbox_fields};
+                my $instance             = $layout->{instance_id}
```
abeverley, line 2907: Should be `$layout->instance_id`

lib/GADS.pm
```diff
@@ -2753,6 +2755,290 @@ prefix '/:layout_name' => sub {
...
+                  if $layout && $layout->{instance_id};
+
+                error "NO INSTANCE" if !$instance;
+
+                my $checkbox_fields = [ split( ',', $checkbox_fields_full ) ];
```
abeverley, line 2912: See comments above re using `->get_all`

lib/GADS.pm
```diff
@@ -2753,6 +2755,290 @@ prefix '/:layout_name' => sub {
...
+                });
+
+                my $lo = param 'layout_name';
+                return forwardHome({success=>"Report updated"}, "$lo/report");
+            }
+
+            my $user   = logged_in_user;
+            my $layout = var('layout') or pass;
+
+            if ( app->has_hook('plugin.linkspace.data_before_request') ) {
```
abeverley, line 2935: Needed?

lib/GADS.pm
```diff
@@ -2753,6 +2755,290 @@ prefix '/:layout_name' => sub {
...
+
+            my $fields = [
+                map $_->presentation(
+                    group            => $records->is_group,
+                    group_col_ids    => $records->group_col_ids,
+                    sort             => $records->sort_first,
+                    query_parameters => query_parameters,
+                ),
+                uniq @columns
+            ];
```
abeverley, line 2958: Ditto above, should be able to use `->all(user_can_read)`

lib/GADS/Schema/Result/Report.pm
```diff
@@ -0,0 +1,448 @@
...
+=head1 NAME
+
+GADS::Schema::Result::Report
+
+=cut
+
+use strict;
+use warnings;
+
+use Data::Dumper;
```
abeverley, line 14: Needed?

lib/GADS/Schema/Result/Report.pm
```diff
@@ -0,0 +1,448 @@
...
+    my $layouts     = $self->report_layouts;
+
+    error "No name given" unless $name;
+    # panic "No instance to link to" unless $instance_id; #This shouldn't happen, but I'm still going to check for it!
+    error "You must provide at least one row to display in the report" unless $layouts;
+
+    0;
+}
+
+#Will return 1 if the report is new, 0 if it is not - this is done via the ID being 0 for a new report.
```
abeverley, line 223: Should be able to use `in_storage` instead

lib/GADS/Schema/Result/Report.pm
```diff
@@ -0,0 +1,448 @@
...
+
+=cut
+
+sub update_report {
+    my ( $self, $args ) = @_;
+
+    my $guard = $self->schema->txn_scope_guard;
+
+    $self->update( { name => $args->{name} } )
+      if $args->{name} && $args->{name} ne $self->name;
```
abeverley, line 345: Don't need the `ne` - DBIx::Class works this out automatically

lib/GADS/Schema/Result/Report.pm
```diff
@@ -0,0 +1,448 @@
...
+    while ( my $layout = $layouts->next ) {
+        push( @{$report_layouts}, $layout->layout_id );
+    }
+
+    #we grep for less writes
+    foreach my $layout (@$report_layouts) {
+        $self->schema->resultset('ReportLayout')
+          ->find( { report_id => $self->id, layout_id => $layout } )->delete
+          if !grep { $_ == $layout } @{ $args->{layouts} };
+    }
```
abeverley, line 361: Better to avoid use of grep as it can be quite slow. A better methodology would be to find_or_create (or use a hash), and then use a single DBIx::Class query to do a search of any IDs no longer needed and then to a delete on that. An example is here: https://github.com/ctrlo/GADS/blob/ecc499f05774aedb42d32237bd0edf2b0f23fe4e/lib/GADS/Schema/Result/User.pm#L742

lib/GADS/Schema/ResultSet/Report.pm
```diff
@@ -0,0 +1,130 @@
...
+    my ( $self, $id, $record_id ) = @_;
+
+    my $schema = $self->result_source->schema;
+
+    error "Invalid report id provided"
+      unless $id && $id =~ /^\d+$/;
+
+    my $result = $self->find( { id => $id }, { prefetch => 'report_layouts' } )
+      or error "No report found for id $id";
+    $result->schema($schema) if !$result->schema;
```
abeverley, line 30: Not sure the above is needed?

lib/GADS/Schema/ResultSet/Report.pm
```diff
@@ -0,0 +1,130 @@
...
+sub create_report {
+    my ( $self, $args ) = @_;
+
+    my $schema = $self->result_source->schema;
+
+    my $guard = $schema->txn_scope_guard;
+
+    my $layouts = [ map { { layout_id => $_ } } @{ $args->{layouts} } ];
+
+    print STDOUT Dumper $layouts;
```
abeverley, line 108: Remove debug

lib/GADS.pm
droberts-ctrlo: I don't know if I'm honest, it's in most other controller requests, and I don't know what it does (if I'm honest), so I left it in where it was and made changes around it! :grimacing:

lib/GADS/Schema/ResultSet/Report.pm
droberts-ctrlo: It is needed to give the linked schema to the Report object, otherwise it is undef.

droberts-ctrlo:
This should be complete now.

views/reports/edit.tt
```diff
@@ -0,0 +1,81 @@
...
+            %]
+            <div class="hidden alert__no__select" aria-hidden="true">
+              <p class="alert__message">Please select at least one item to add to the report!</p>
+            </div>
+            <fieldset class="fieldset--required fieldset--report" name="all_checkboxes">
+              [% FOREACH field IN fields %]
+              <input [%IF field.is_checked%]checked [% END %]type="checkbox" id="[% field.id %]" name="checkboxes" value="[% field.id %]" class="checkbox__input" />
+              <label for="[% field.id %]" class="checkbox__label">[% field.name %]</label><br />
+              [% END %]
+            </fieldset>
```
abeverley, line 54: I've tried using the existing snippets, and it seems to work with the following code:

```
[%
  report_items = [];

  FOREACH field IN fields;
    report_items.push({
      field_id   = field.id,  
      name       = field.name,
      id         = field.id,  
      is_checked = field.is_checked ? 1 : 0
    });
  END;

  INCLUDE fields/checkbox_list.tt
    list_class        = "list--vertical list--checkboxes"
    name              = "checkboxes" 
    label             = "Columns to show in report:"
    items             = report_items 
    fieldset_required = 1
    fieldset_classes  = "fieldset--report"
    filter            = "html";
%]
```

src/frontend/components/button/lib/create-report-button.js
```diff
@@ -0,0 +1,82 @@
...
+  initSubmitReport() {
+    this.el.on('click', (ev) => { this.submitReport(ev) });
+  }
+
+  /**
+   * Check if at least one checkbox is checked
+   * @param {jQuery<HTMLFieldSetElement>} $fieldset fieldset to check
+   * @returns {boolean} true if at least one checkbox is checked
+   */
+  checkForAtLeastOneValue($fieldset) {
```
abeverley, line 33: For this, I'm wondering whether it's possible to reuse the code in `src/frontend/js/lib/validation.js` (e.g. the `validateCheckboxGroup` function)

lib/GADS.pm
```diff
@@ -2753,6 +2753,235 @@ prefix '/:layout_name' => sub {
...
+            my $user   = logged_in_user;
+            my $layout = var('layout') or pass;
+
+            my %params = (
+                user   => $user,
+                layout => $layout,
+                schema => schema,
+            );
+
+            my $records = GADS::Records->new(%params);
```
abeverley, line 2769: I think this code can be removed now?

lib/GADS.pm
```diff
@@ -2753,6 +2753,235 @@ prefix '/:layout_name' => sub {
...
+                user   => $user,
+                layout => $layout,
+                schema => schema,
+            );
+
+            my $base_url = request->base;
+
+            my $params;
+
+            $params->{alerts}      = $alert->all;
```
abeverley, line 2781: I'm not sure that $alerts is needed?

lib/GADS.pm
```diff
@@ -2753,6 +2753,235 @@ prefix '/:layout_name' => sub {
...
+            $params->{layout_obj} = $layout;
+            $params->{layout}     = $layout;
+
+            $params->{header_back_url} = "${base_url}table";
+            $params->{breadcrumbs}     = [
+                Crumb( $base_url . "table/", "Tables" ),
+                Crumb( "",                   "Table: " . $layout->name )
+            ];
+
+            my $layout_id = $layout->{instance_id};
```
abeverley, line 2793: I don't think this is used?

lib/GADS.pm
```diff
@@ -2753,6 +2753,235 @@ prefix '/:layout_name' => sub {
...
+
+            my $base_url = request->base;
+
+            my $params;
+
+            $params->{alerts}      = $alert->all;
+            $params->{header_type} = 'table_tabs';
+
+            $params->{layout_obj} = $layout;
+            $params->{layout}     = $layout;
```
abeverley, line 2785: I don't think $params->{layout} is used

lib/GADS.pm
```diff
@@ -2753,6 +2753,235 @@ prefix '/:layout_name' => sub {
...
+            template 'reports/view' => $params;
+        };
+
+        #add a report
+        any [ 'get', 'post' ] => '/add' => require_login sub {
+            if ( body_parameters && body_parameters->{submit} ) {
+                my $layout             = var('layout') or pass;
+                my $report_description = params->{'report_description'};
+                my $report_name        = params->{'report_name'};
+                my $checkbox_fields    = params->{'checkboxes'};
```
abeverley, line 2808: Can you change `params` to `body_parameters->get()` (or `get_all`) please? Using body_parameters ensures that URL query parameters are not used, and it also ensures a consistent type of variable (e.g. if you try submitting a report with only one field ticked you will get an error, as the array becomes a scalar without the use of `get_all`

lib/GADS.pm
```diff
@@ -2753,6 +2753,235 @@ prefix '/:layout_name' => sub {
...
+                my $instance           = $layout->{instance_id}
+                  if $layout && $layout->{instance_id};
```
abeverley, line 2810: Please access instance_id as a method (i.e. `$layout->instance_id`) as using a hash will not always work correctly. Also, I don't think there needs to be an `if` here as `$layout` will always be true.

lib/GADS.pm
```diff
@@ -2753,6 +2753,235 @@ prefix '/:layout_name' => sub {
...
+
+                my $user = logged_in_user;
+
+                my $report = schema->resultset('Report')->create_report(
+                    {
+                        schema      => schema,
```
abeverley, line 2816: I don't think schema is needed

lib/GADS.pm
```diff
@@ -2753,6 +2753,235 @@ prefix '/:layout_name' => sub {
...
+                    }
+                );
+
+                my $lo = param 'layout_name';
+                return forwardHome( { success => "Report created" },
+                    "$lo/report" );
+            }
+
+            my $user   = logged_in_user;
+            my $layout = var('layout') or pass;
```
abeverley, line 2832: These variables can be initialised before the submit check so they are only needed once.

lib/GADS.pm
```diff
@@ -2753,6 +2753,235 @@ prefix '/:layout_name' => sub {
...
+
+            if ( app->has_hook('plugin.linkspace.data_before_request') ) {
+                app->execute_hook( 'plugin.linkspace.data_before_request',
+                    user => $user );
+            }
```
abeverley, line 2837: Hook can be removed

lib/GADS.pm
```diff
@@ -2753,6 +2753,235 @@ prefix '/:layout_name' => sub {
...
+
+            my %params = (
+                user   => $user,
+                layout => $layout,
+                schema => schema,
+            );
```
abeverley, line 2843: I think this can be removed.

lib/GADS.pm
```diff
@@ -2753,6 +2753,235 @@ prefix '/:layout_name' => sub {
...
+            ];
+
+            $params->{viewtype} = 'add';
+            $params->{fields}   = $records;
+
+            template 'reports/edit' => $params;
+        };
+
+        #Edit a report (by :id)
+        any [ 'get', 'post' ] => '/edit:id' => require_login sub {
```
abeverley, line 2869: I think most comments in the previous route also apply to this route. In the main I think it can be collapsed down to be a lot shorter with some of the code removed.

lib/GADS.pm
```diff
@@ -2753,6 +2753,235 @@ prefix '/:layout_name' => sub {
...
+            my $lo = param 'layout_name';
+            return forwardHome( { success => "Report deleted" }, "$lo/report" );
+        };
+
+        #Render the report (by :report) with the view (by :view)
+        get "/render:report/:view" => sub {
+            my $user = logged_in_user;
+
+            my $report_id = param('report');
+            my $view_id   = param('view');
```
abeverley, line 2974: These should be `route_parameters` (I realised a lot of the old code uses `param` but it's best to move away from that for new code)

lib/GADS.pm
```diff
@@ -2753,6 +2753,235 @@ prefix '/:layout_name' => sub {
...
+
+            my $report =
+              schema->resultset('Report')->load( $report_id, $view_id );
```
abeverley, line 2977: Thinking about this a bit more, I think a nicer way of implementing it would be to remove `load()` to `Result::Report` and then use `find`. E.g. `my $report = schema->resultset('Report')->find($report_id)->load($view_id)`

lib/GADS/Schema/ResultSet/Report.pm
```diff
@@ -0,0 +1,128 @@
...
+sub load_all_reports {
+
+    my ( $self, $instance_id, $schema ) = @_;
+
+    error "Invalid layout provided"
+      unless $instance_id && $instance_id =~ /^\d+$/;
+
+    my $items = $self->search(
+        {
+            instance_id => $instance_id,
```
abeverley, line 76: `deleted` could be added to the search here

lib/GADS/Schema/ResultSet/Report.pm
```diff
@@ -0,0 +1,128 @@
...
+        },
+        {
+            prefetch => 'report_layouts',
+        }
+    );
+
+    my $result = [];
+
+    while ( my $next = $items->next ) {
+        $next->schema($schema)    if !$next->schema;
```
abeverley, line 86: Adding schema to the object shouldn't be needed?

abeverley (commented):
Thanks Dave, this is really good. Sorry for all the comments! There may be a few more things once I've cast my eyes over it again, but this review should cover most things. Many thanks.

src/frontend/components/button/lib/create-report-button.js
droberts-ctrlo: No problem, the main reason this is here is I didn't want to "fiddle" with the code that's already working and in place for validation (and it's easier to test here) - I'll look at moving it over.

lib/GADS.pm
droberts-ctrlo: The if was there more as a "just making sure I didn't break it" statement - I like to make sure I've not missed anything. I'll remove :-)

lib/GADS/Schema/ResultSet/Report.pm
droberts-ctrlo: I looked into this (admittedly only using `print` statements), and the schema object doesn't "attach" automatically, so I need to attach it here - the if statement was there as part of that check, and I'll remove it.

abeverley: I think you should be able to access it from within the object by using something like `$self->result_source->schema` rather than needing to pass it in, which means you could also get rid of the `schema` attribute on the object.

droberts-ctrlo:
That's the new iteration finished (I believe) - :+1:

abeverley (commented):
Perl comments as discussed.

lib/GADS.pm
```diff
@@ -2753,6 +2761,205 @@ prefix '/:layout_name' => sub {
...
+
+            template 'reports/edit' => $params;
+        };
+
+        #Edit a report (by :id)
+        any [ 'get', 'post' ] => '/edit:id' => require_login sub {
+
+            my $user   = logged_in_user;
+            my $layout = var('layout') or pass;
+            my $report_id = param('id');
```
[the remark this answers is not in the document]
droberts-ctrlo: Not permickety, neat! :)

lib/GADS/Schema/Result/Report.pm
```diff
@@ -0,0 +1,423 @@
...
+sub validate {
+    my ( $self, $value, %options ) = @_;
+
+    my $name        = $self->name;
+    my $instance_id = $self->instance_id;
+    my $layouts     = $self->report_layouts;
+
+    error "No name given" unless $name;
+    error "You must provide at least one row to display in the report"
+      unless $layouts;
```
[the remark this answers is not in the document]
droberts-ctrlo: It appeared to in testing - although I'm not sure how to mitigate if it doesnt as at least one layout is required to "properly" create a report (I changed the save code to include layouts as required) ?

lib/GADS/Layout.pm
```diff
@@ -295,6 +295,25 @@ has _user_permissions_columns => (
...
+    my $reports_rs = $self->schema->resultset('Report')->search({
+        instance_id => $self->instance_id,
+        deleted => undef
+    });
+    my @reports;
+    while (my $report = $reports_rs->next)
+    {
+        push @reports, $report;
+    }
+    \@reports;
```
abeverley, line 314: The whole of the above can now be written as: `[$report_rs->all]`

lib/GADS/Schema/ResultSet/Report.pm
```diff
@@ -0,0 +1,115 @@
...
+        }
+    );
+
+    my $result = [];
+
+    while ( my $next = $items->next ) {
+        push( @{$result}, $next );
+    }
+
+    return $result;
```
abeverley, line 80: The whole of the above can be rewritten as `[$items->all]`

PR #262, opened 2023-11-02
Added Remove Alert button

Oliver-ctrlo:
- Removed 'Never option from the alert frequency options.

- Modified some of the displayed text for the alert modal header and buttons. 

- Added A button called 'Remove' to the alert modal that will replace the 'Never' option in the Alert drop-down.

- The 'Remove' button functions in exactly the same way as the create button but instead it's fixed to always write the 'Never' value in the alert frequency. (as well as having its own return message).

### DIFF
--- a/lib/GADS.pm
+++ b/lib/GADS.pm
@@ -2296,20 +2296,28 @@ prefix '/:layout_name' => sub {
         }
 
         # Deal with any alert requests
-        if (param 'modal_alert')
-        {
+        if (param('modal_alert') || param('modal_remove')) {
+            my $success_message;
+            my $frequency = '';
+    
+            if (param('modal_remove')) {
+                $frequency = '';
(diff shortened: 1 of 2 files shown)


lib/GADS.pm
```diff
@@ -2296,20 +2296,28 @@ prefix '/:layout_name' => sub {
...
 
         # Deal with any alert requests
-        if (param 'modal_alert')
-        {
+        if (param('modal_alert') || param('modal_remove')) {
+            my $success_message;
+            my $frequency = '';
+    
+            if (param('modal_remove')) {
+                $frequency = '';
```
abeverley, line 2304: Frequency is defined as an empty string twice, so one could be removed.

abeverley:
Thanks Oliver. Just a minor comment but I'll merge anyway.

PR #263, opened 2023-11-02
View Filter Changes

droberts-ctrlo:
Menu below view filters now complete using textbox


droberts-ctrlo:
This is _not_ complete - Have discovered a bug that needs fixing!

droberts-ctrlo:
This should now be ready for review

PR #271, opened 2023-11-13
Spinner should be complete

droberts-ctrlo:
This requires testing with larger datasets


abeverley:
Overtaken by events, already fixed with commit 95b1223720ba0e27c5c28054e8b6c784847420ae

PR #274, opened 2023-11-14
SECURITY FIX

droberts-ctrlo:
Users could view all records within the report regardless of permissions, this is now fixed


lib/GADS/Schema/Result/Report.pm
```diff
@@ -220,18 +220,32 @@ has data => (
 
         my $record_id = $self->record_id;
         $record->find_current_id($record_id);
+        my $allowed_cols = [map {$_->name} $gads_layout->all(user_can_read => 1)];
 
-        while ( my $layout = $layouts->next ) {
+         while ( my $layout = $layouts->next ) {
+            next unless _is_column_allowed($self, $layout->layout->name, $allowed_cols);
             my $column = $gads_layout->column( $layout->layout_id );
```
abeverley, line 227: Thanks Dave. A simpler fix might be just to change the above line to:
`my $column = $gads_layout->column( $layout->layout_id, read => 1 ) or next`
Can you try that please?

abeverley: (also, skipping rather than showing blank would make it consistent with other parts of the app)

droberts-ctrlo: Doesn't work - can now read values I'm not allowed to

droberts-ctrlo: And it does skip :smile:

droberts-ctrlo, line 227: Although I did forget a $self there to make it consistent

droberts-ctrlo: Fixed this using

`my $column = $gads_layout->column( $layout->layout_id, permission=>'read' );`

lib/GADS/Schema/Result/Report.pm
```diff
@@ -221,10 +221,10 @@ has data => (
         my $record_id = $self->record_id;
         $record->find_current_id($record_id);
 
-        while ( my $layout = $layouts->next ) {
-            my $column = $gads_layout->column( $layout->layout_id );
+         while ( my $layout = $layouts->next ) {
+            my $column = $gads_layout->column( $layout->layout_id, permission=>'read' ) or next;
```
abeverley, line 225: Thanks. Just fussy white space now - `while` has gained a space in front of it, and just personal preference but `permission => 'read'` keeps that consistent too :)

abeverley:
> Also don't know why the other files are showing as edited - I pulled and used the latest UIUX for this. This is a lovely GitHub issue, methinks!

Yes, it does appear to be a "github-ism". Doing through the command line, `git diff uiux..droberts/reporting-fix` will show as expected, `git diff uiux...droberts/reporting-fix` appears to be what Github is using.

PR #275, opened 2023-11-14
Validation fix

droberts-ctrlo:
Fix for validation errors on fieldsets


droberts-ctrlo:
Fully tested (manual) and found to be working as expected in local test environment

PR #276, opened 2023-11-15
Searchableuserfields

droberts-ctrlo:


droberts-ctrlo:
Now ready to merge - had some odd file changes, but these have been removed

abeverley:
Thanks Dave, this is great. It's good to merge, but just some minor feedback for general info:

- I would write the $titles line as:
```
my @titles = map {
  +{ label_html => $title->name, value => $title->id }
} @{$userso->titles};
```

- I'd prefer less in the controller, although that's more a problem with the general code layout than this particular commit.

abeverley:
Also, can you just confirm that the HTML entities are escaped as expected please? E.g. a title such as `My <b>title</b>` does not get rendered as HTML.

PR #277, opened 2023-11-17
Reporting fixes

droberts-ctrlo:
- Added delete modal
- Moved create report button to different place for consistency with other views
- Changed the button from Reports to Download options when viewing records
- Fixed overflow displaying off edge of screen for long report names
- Removed ability to view/edit deleted reports
- Fixed error on button code
- Removed extra unneeded code
- Created utility class for functions that are being used across a number of features (this could have been another request, but I keep having to make this file with most of what I'm doing every time, so it's better to push it now in my opinion)


abeverley:
Thanks Dave, I just made some minor formatting tweaks in f9f41aac (worth having a quick check) - I just didn't want to hold things up.

Also, I think the PR had an incorrect code reversion which I have fixed in 47d670ca - can you double-check please? (And also make sure nothing else incorrect was committed by looking at d224119b

PR #280, opened 2023-11-20
Fixes for errors found in testing

droberts-ctrlo:


src/frontend/test/validation.js
```diff
@@ -159,7 +159,9 @@ const validateRequiredFields = (form) => {
 
 const validateRequiredFieldsetCheckboxes = (fieldset) => {
   let isValid = false;
-  fieldset.find("input[type=checkbox]").each((i, field) => {
+  const checkboxes = fieldset.find("input[type=checkbox]");
+  if(checkboxes.length === 0) return true;
+  checkboxes.each((i, field) => {
```
abeverley, line 164: Not sure if the above is intentional?

views/fields/select_single.tt
```diff
@@ -76,7 +76,7 @@
                   data-value="[% item.label_html %]"
               >
               <label class="radio-group__label" for="[% id %]_[% item.value %]">
-                [% INCLUDE fields/sub/filter.tt value=item.label_html filter=filter %]
+                [% INCLUDE fields/sub/filter.tt value=item.label_html filter="html" %]
```
abeverley, line 79: I've been looking a bit more at this, and I think it needs altering slightly. At the moment, the `select_single` expects `items` containing both `label_plain` and `label_html` keys (the latter being HTML entities encoded). I think adding the filter `html` as above will result in double-encoding when the widget is used for editing records.

I think you had the right idea originally, passing in the `filter` parameter. I suspect that `filter` is a reserved keyword though, so it would be worth trying again by renaming it (and also updating `edit.tt` which seems to have the wrong filter set already). Then it will also need `select_single.tt` updating to allow only only `label_plain` being provided in the case of `filter="html"` and automatically filling `label_html` in this case.

Hope that makes sense, if not can discuss next time I'm in!

lib/GADS.pm
```diff
@@ -1606,14 +1606,18 @@ any ['get', 'post'] => '/user/:id' => require_any_role [qw/useradmin superadmin/
 
     my $orgs = [map { +{ label_html => $_->name, value => $_->id } } @{$userso->organisations}];
 
+    my $depts = [map {+{label_html=>$_->name, value=>$_->id}} @{$userso->departments}];
```
abeverley, line 1609: See comments below, `label_html` is expected to be HTML-encoded.

abeverley (changes requested):
Thanks Dave. As always, things are never easy - hope my comments make sense!

src/frontend/test/validation.js
droberts-ctrlo: I'll be honest, we've hit "that point" we were discussing previously about having the changes to the files in 2 places at the same time - it's becoming difficult to keep track of what's where. I've reverted this one, and it needs to be kept "as is". :smile:

views/fields/select_single.tt
```diff
@@ -76,7 +76,15 @@
...
-                [% INCLUDE fields/sub/filter.tt value=item.label_html filter=filter %]
+                [%
+                    IF item.label_plain;
+                        INCLUDE fields/sub/filter.tt value=item.label_plain filter=text_filter
+                    ELSIF item.label_html;
+                        item.label_html;
+                    ELSE;
+                        item;
+                    END;
+                %]
```
abeverley, line 87: A few more changes needed here. It's possible that `item` will contain both `label_plain` and `label_html`. If it does, `label_html` should be used in preference for this particular part of `select_single.tt`. However there are other parts of this template that also need changing accordingly. For example, further up the file there is `data-list-text` which should always use `label_plain` and a `span` which should use `label_html` if available or otherwise HTML-encode `label_plain`.

Having thought more about the `filter` parameter, I think best to remove that entirely from `user_edit.tt` and `edit.tt` and then always filter using `INCLUDE fields/sub/filter.tt value=item.label_plain filter="html"` (here and above).

lib/GADS.pm
```diff
@@ -1602,18 +1602,22 @@ any ['get', 'post'] => '/user/:id' => require_any_role [qw/useradmin superadmin/
...
 
-    my $titles = [map { +{ label_html => $_->name, value => $_->id } } @{$userso->titles}];
+    my $titles = [ map { +{ label_plain => $_->name, value => $_->id } } @{ $userso->titles } ];
 
-    my $orgs = [map { +{ label_html => $_->name, value => $_->id } } @{$userso->organisations}];
+    my $orgs = [ map { +{ label_plain => $_->name, value => $_->id } } @{ $userso->organisations } ];
+
+    my $depts = [ map { +{ label_plain => $_->name, value => $_->id } } @{ $userso->departments } ];
+
+    my $teams = [ map { +{ label_plain => $_->name, value => $_->id } } @{ $userso->teams } ];
```
abeverley, line 1611: Another thing I have just realised is that all of the above also needs to be passed/updated into the `user_overview` template in the `user_overview` route. As such, rather than repeat all the above code, I think it would be better to move the code into the `GADS::Users` class.

droberts-ctrlo: Changed the "build" code for each attribute - although part of me wonders if this will "break" something else(?) If preferred, I'll give each it's own attribute seperate from the current ones

abeverley: Thanks, this is definitely the way to go. However, you are correct, I think it may have broken the "register" and "my account" page. Can you take a look at those please and fix as required please? It looks like it would be worth updating them to use the same select drop-down too.

It looks like the `/login` route also has them in the route but that they are not used in the template, can you check and remove from the controller as required please?

views/fields/select_single.tt
abeverley: I think the first paragraph of the comment above still needs resolving.

lib/GADS.pm
droberts-ctrlo: Completed as above

PR #281, opened 2023-11-20
Typeahead Upgrade

droberts-ctrlo:
- Changed currently used typeahead library for one with better documentation (that is also more current)
- Retrofitted to areas where old typeahead was used
- Created abstraction for use within system
    - Typeahead class created
    - TypeaheadOptions interface created
    - TypeaheadBuilder class created

Abstraction was done in order to make code easier to read and the possibility of controlling typeaheads as a single unit (thus keeping behaviour consistent). The builder was placed in order to make construction of a typeahead more verbose in order to mitigate errors within the configuration.


droberts-ctrlo:
This should be ready to merge

droberts-ctrlo:
This should now be fully ready to merge

droberts-ctrlo:
With a view to merging current release (going to EPF) and do branch (and merge) changes.

PR #282, opened 2023-11-22
Changed calc type from input to auto adjust textarea

Oliver-ctrlo:
- Changed the field type of calc from input to textarea.
- Changed the re-calc triggers to point at text-area instead of input.
- Added class + functionality to calc textareas,  to allow text-areas to resize based on the content returned by the Calc. (on load and on change).

### DIFF
--- a/src/frontend/components/form-group/calc-fields/lib/component.js
+++ b/src/frontend/components/form-group/calc-fields/lib/component.js
@@ -62,10 +62,12 @@ class CalcFieldsComponent extends Component {
         var returnval = func.apply(first, vars)
 
         // Update the field holding the code's value
-        $field.find('input').val(returnval)
+        $field.find('textarea').val(returnval)
+        $field.find('textarea').text(returnval)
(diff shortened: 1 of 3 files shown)


src/frontend/components/form-group/calc-fields/lib/component.js
```diff
@@ -62,10 +62,12 @@ class CalcFieldsComponent extends Component {
         var returnval = func.apply(first, vars)
 
         // Update the field holding the code's value
-        $field.find('input').val(returnval)
+        $field.find('textarea').val(returnval)
```
abeverley, line 65: Is `.val()` definitely needed here?

abeverley:
This is great, thanks Oliver. Please just check the above comment before I merge.

src/frontend/components/form-group/calc-fields/lib/component.js
Oliver-ctrlo: Correct it's no longer needed. Now Removed.

PR #286, opened 2023-11-27
Found user changes "broke" things

droberts-ctrlo:
Import user, among others, stopped working due to changes - adding users wizard is no longer allowing addition of depts, titles, etc. This now works, but has gone back to previous functionality. It shall be changed in the next pull request.


abeverley:
Thanks Dave, good shout. As it is, I think I've been able to fix the problems (@oliver-ctrlo to confirm), so I'll close this for the time being.

PR #288, opened 2023-11-30
HOTFIX - Ticket 1484

droberts-ctrlo:


droberts-ctrlo:
Would this not be redundant due to Line 81 checking fieldsets anyway - if we include a check in there to ignore all fieldsets, we'll lose the changes to text inputs. Would it not be better to change it so that it ignores the fieldsets, and just does the check you've suggested above at line 66 (i.e. `if($(field).prop('type') === 'radio' || $(field).prop('type') === 'checkbox' ) {...` and disregard fieldsets altogether - whichever way you look at it, the code from line 80 onwards loops all input controls anyway, so removing 59-77 and placing the functionality below would probably reduce the complexity and speed things up (albeit marginally).

droberts-ctrlo:
> Would this not be redundant due to Line 81 checking fieldsets anyway - if we include a check in there to ignore all fieldsets, we'll lose the changes to text inputs. Would it not be better to change it so that it ignores the fieldsets, and just does the check you've suggested above at line 66 (i.e. `if($(field).prop('type') === 'radio' || $(field).prop('type') === 'checkbox' ) {...` and disregard fieldsets altogether - whichever way you look at it, the code from line 80 onwards loops all input controls anyway, so removing 59-77 and placing the functionality below would probably reduce the complexity and speed things up (albeit marginally).

Thinking about it - the checkboxes would need to be within a "push" into an array, not repeated values, so this may not be the best way - although I think the theory is sound!

PR #292, opened 2023-11-30
Typeahead upgraded

droberts-ctrlo:
Redid commit as there were too many changes in the previous commit that were erroneous.


.babelrc
droberts-ctrlo, line 1: Found that this works without issue, have removed lines to allow for removal of test directory

src/frontend/js/lib/util/mapper/mapper.ts
droberts-ctrlo, line 1: Basic mapper for AJAX requests

src/frontend/js/lib/util/typeahead/lib/TypeaheadBuilder.test.ts
droberts-ctrlo, line 1: *Very* basic tests

src/frontend/js/lib/validation.js
droberts-ctrlo, line 1: Validation needs below change to work with tests - this does not affect functionality

tsconfig.json
droberts-ctrlo, line 1: As I am now using TypeScript, a "fuller" configuration is required - this does not affect WebPack, just testing.

droberts-ctrlo:
Tested on local system by OT - all appears working as expected

PR #294, opened 2023-11-30
Datatables typeahead

droberts-ctrlo:
Require merge of #292 first


src/frontend/components/data-table/lib/component.js
```diff
@@ -1,15 +1,14 @@
-import { Component } from 'component'
+import { Component, initializeRegisteredComponents } from 'component'
```
droberts-ctrlo, line 1: Minor cleanup - code was "wrong" but not an error

lib/GADS/API.pm
```diff
@@ -546,6 +546,27 @@ any ['get', 'post'] => '/api/:sheet/records' => require_login sub {
     _get_records();
 };
 
+get '/api/:sheet/fields' => require_login sub {
+    my $user             = logged_in_user;
+    my $sheetname        = param 'sheet'
+        or panic __"No sheet name found";
```
abeverley, line 552: This should probably be `error`, as it is more likely to be a user error than a coding error

lib/GADS/API.pm
```diff
@@ -546,6 +546,27 @@ any ['get', 'post'] => '/api/:sheet/records' => require_login sub {
...
+    my $layout           = var('instances')->layout_by_shortname($sheetname)
+        or panic __x"Sheet {name} not found", name => $sheetname;
+    my $col_title        = query_parameters->get('title')
+        or panic __"No column title found";
```
abeverley, line 556: Same for this 2 as well.

lib/GADS/API.pm
```diff
@@ -546,6 +546,27 @@ any ['get', 'post'] => '/api/:sheet/records' => require_login sub {
...
+    my $search           = query_parameters->get('search');
+
+    my @columns = $layout->all(user_can_read => 1);
+    #Not sure if I should use the below in case it returns the column regardless of permissions?
+    #my $column = $layout->column_by_name($col_title);
```
abeverley, line 561: It would be better to use the column ID instead, as that is guaranteed to be unique. You could then do:
```
my $column = $layout->column($col_id, permission => "read")
    or error __x"Column {id} not found", id => $col_id;
```

src/frontend/components/data-table/lib/component.js
```diff
@@ -278,11 +276,39 @@ class DataTableComponent extends Component {
 
     this.toggleFilter(column)
 
+    let setField = title.trim();
+    if (setField.endsWith('Sort')) {
+      setField = setField.substring(0, setField.length - 4);
+    }
```
abeverley, line 282: See previous comments about column ID. It should be possible to get this with something like `.oAjaxData.columns[index].name` (see examples in `src/frontend/components/data-table/lib/component.js`)

src/frontend/components/data-table/lib/component.js
```diff
@@ -323,6 +349,11 @@ class DataTableComponent extends Component {
     })
   }
 
+  getApiEndpoint(title) {
+    const table = window.location.pathname.substring(1).split('/')[0];
+    return `/api/${table}/fields?title=${title}&search=`;
```
abeverley, line 354: There should be a better way of getting the table name. One option might be to use `$(body).data("layout-identifier")`

abeverley (commented):
Thanks Dave, looking really good. I've added a few minor comments. One thing I noticed is that it doesn't seem to work on tree columns (`Uncaught TypeError: e.name.toLowerCase is not a function`). Can you check against all column types please?

abeverley:
Actually, I've just realised, instead of the new endpoint in API.pm, I think you should be able to use this (once updated to use column ID instead of name): https://github.com/ctrlo/GADS/blob/57e218e084221e4afc6e62ce87b4d50b838dee5f/lib/GADS.pm#L4275

src/frontend/components/data-table/lib/component.js
droberts-ctrlo: I didn't even spot that in the code, yet another useful titbit to add to my log! :smile:

PR #297, opened 2023-12-01
Addition of user notes to account requests

droberts-ctrlo:


views/user/user_request.tt
```diff
@@ -109,6 +113,12 @@
         filter = "html_entity"
         name   = "firstname"
         value  = user_row.firstname
+      }, {
+        type  = "link"
+        label = user_row.account_request_notes
+        filter = "html_entity"
+        name  = "notes"
+        value = user_row.account_request_notes
```
abeverley, line 121: Thanks Dave, can you just line up the "equals" signs please? (I know not all files are aligned, but it would be good to at least do the ones above that already are). Thanks!

PR #299, opened 2023-12-05
Fix for welcome email not being sent on approval of account

droberts-ctrlo:


abeverley:
Thanks Dave. Looking at this in a bit more detail, I think it's probably best instead to combine `_post_add_user_account` and `_post_request_account` if you think that's possible? That will prevent any similar bugs in the future, with code being duplicated.

lib/GADS/Schema/Result/User.pm
```diff
@@ -965,6 +967,27 @@ sub update_user
...
+    if(!!$request) {
+      $self->send_welcome_email($params{email});
+    }
+}
+
+sub send_welcome_email
+{   my ($self, %params) = @_;
+
+    $params{email} = $self->email
+      unless defined $params{email} || $params{email};
```
abeverley, line 979: Could this be simplified as follows?
```
$params{email} ||= $self->email
```

lib/GADS/Schema/Result/User.pm
```diff
@@ -965,6 +967,27 @@ sub update_user
 
     $guard->commit;
 
+    if(!!$request) {
+      $self->send_welcome_email($params{email});
+    }
```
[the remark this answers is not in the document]
droberts-ctrlo: Done.

lib/GADS/Schema/Result/User.pm
droberts-ctrlo: Didn't know you could do that!

lib/GADS/Schema/ResultSet/User.pm
```diff
@@ -255,7 +240,7 @@ sub upload
     # Won't get this far if we have any errors in the previous statement
     $guard->commit;
 
-    $self->_send_welcome_email(%$_, request_base => $options{request_base})
+    Schema::Result::User::send_welcome_email(%$_, request_base => $options{request_base})
         foreach @welcome_emails;
```
abeverley, line 244: Sorry to be a pain... looking at this afresh, I think that `@welcome_emails` should be changed to be an array of `Schema::Result::User` objects (i.e. change above to be `push @welcome_emails, $u`) and then the above could be written as:
```
$_->send_welcome_email(code => $_->resetpw, request_base => $options{request_base})
    foreach @welcome_emails;
```
It's basically the same, I just think it's generally better to call the method `send_welcome_email` method natively.

Or something like that - it would need testing.

droberts-ctrlo: Ironically, I was thinking the same a moment ago, then I thought I'd see if you thought the same (you did).

PR #303, opened 2023-12-11
Account request notes fix

droberts-ctrlo:
There were 2 different fields within account request notes for "Notes" - it looks like one of them was added in error (as the "user" class has no "notes" field). As such, I have removed the "extra" one and fixed the naming and settings of the field that was producing an error.


droberts-ctrlo:
Hi Andy. I'm inclined to agree there, changed as suggested.

PR #304, opened 2023-12-12
Fixed inability to remove title, department, team, and organisation

droberts-ctrlo:


lib/GADS/Schema/Result/User.pm
```diff
@@ -852,14 +852,14 @@ sub update_user
...
-    delete $params{team_id} if !$params{team_id};
-    delete $params{title} if !$params{title};
-
     my $site = $self->result_source->schema->resultset('Site')->next;
 
+    # Set null values where required for database insertions
+    delete $params{organisation} if !$params{organisation} && !$site->user_field_is_editable('organisation');
+    delete $params{department_id} if !$params{department_id} && !$site->user_field_is_editable('department_id');
+    delete $params{team_id} if !$params{team_id} && !$site->user_field_is_editable('team_id');
+    delete $params{title} if !$params{title} && !$site->user_field_is_editable('title');
```
abeverley, line 861: Thanks for the explanation Dave. I'm wondering instead though whether the following would be better?
```
$params{organisation} ||= undef;
```
I've not tried this, but I'm thinking that dealing with blank values should be unrelated to whether a field is editable? I think the original code was in there to stop empty strings trying to be inserted into integer fields of the user table.

droberts-ctrlo: This unfortunately wouldn't work, because if the field isn't editable, it isn't passed to the sub. In this case, that would mean that on saving, the field would clear every time if the field wasn't user-editable.

PR #307, opened 2023-12-13
Second iteration of reporting updates

droberts-ctrlo:
Relies on #306

- Added logo upload
- Added logo to PDF
- Added heading spacing
- Added multi-line for headings

PR #308, opened 2023-12-15
Reporting updates 3

droberts-ctrlo:
Grouping by topic now possible

Requires #307 merging first


droberts-ctrlo:
I think I'm overcomplicating this - closing all requests and working from there!

PR #314, opened 2024-01-09
Fix for fields not displaying

droberts-ctrlo:
Notes field on User Wizard had "vanished" for some reason. This has been added back in!


abeverley:
Thanks Dave. Looking at the previous changes, the notes field was added into `firstFrameFields` whereas this is for `secondFrameFields`. Looking at the code the notes field should display in the former. That's not to say that I can't merge this, I'm just wondering whether there is something else going on that makes the notes field drop into the second frame?

PR #320, opened 2024-01-11
Typeahead rag fix

droberts-ctrlo:
- Added code to display RAG according to the inputted value on enabled_rags of the layout
- Removed typeahead from date fields as this would over-complicate functionality

### DIFF
--- a/src/frontend/components/data-table/lib/component.js
+++ b/src/frontend/components/data-table/lib/component.js
@@ -247,6 +247,7 @@ class DataTableComponent extends Component {
     const {oAjaxData} = context[0];
     const {columns} = oAjaxData;
     const columnId = columns[column.index()].name;
+    const columnType = $header.find(".data-table__header-wrapper").data("type");
 
     const $searchElement = $(
       `<div class='data-table__search'>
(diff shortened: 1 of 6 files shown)


abeverley:
Thanks Dave. As discussed please:
- Look at a Moo::Role object to abstract the RAG functions
- Move the RAG values retrieval from the controller into Column::Rag
- Consider moving the "has filter typeahead" test into column classes in the backend

src/frontend/components/data-table/lib/component.js
```diff
@@ -247,6 +249,7 @@ class DataTableComponent extends Component {
     const {oAjaxData} = context[0];
     const {columns} = oAjaxData;
     const columnId = columns[column.index()].name;
+    const col = this.columns.filter((col) => col.name === columnId.replace(/\s+/g, ''))[0];
```
abeverley, line 252: I think the above could be simplified and replaced by:
```
const col = this.columns[column.index()];
```

abeverley:
Thanks Dave, just the one inline comment above, plus I noticed that the search results are appearing in uppercase. This seems to be because of the `text-transform: uppercase` on `.data-table thead th`. I was going to say remove that altogether as it is also covered with `.data-table__sort`; however, the latter would not be present on tables without a sortable column, so maybe find a way to make sure it only applies to the column header itself, not the text in anything attached to it such as the typeahead.

PR #322, opened 2024-01-15
Reporting updates fixed

droberts-ctrlo:


lib/GADS.pm
```diff
@@ -1329,6 +1329,49 @@ any ['get', 'post'] => '/settings/team_edit/:id' => require_any_role [qw/useradm
...
+    my $site             = var 'site';
+
+    if(body_parameters->get('submit')) {
+        my $post_marking = body_parameters->get('security_marking');
+
+        my $txn_scope_guard = schema->txn_scope_guard;
+
+        $site->update({ security_marking => $post_marking });
+
+        $txn_scope_guard->commit;
```
abeverley, line 1342: Doesn't really matter, but just to note that there is no need for a guard if only making one SQL change

lib/GADS/API.pm
```diff
@@ -546,6 +546,40 @@ any ['get', 'post'] => '/api/:sheet/records' => require_login sub {
...
+    if ( my $file = upload('file') ) {
+        #should this not return a json response with error set to 1?
+        forwardHome({ danger => "You do not have permission to manage system settings"}, '')
+            unless logged_in_user->permission->{superadmin};
+
+        my $txn_scope_guard = schema->txn_scope_guard;
+
+        $site->update({ site_logo => $file->content });
+
+        $txn_scope_guard->commit;
```
abeverley, line 561: Same comment regarding guard: no harm, but not necessary either.

lib/GADS/API.pm
```diff
@@ -546,6 +546,40 @@ any ['get', 'post'] => '/api/:sheet/records' => require_login sub {
...
+        content_type 'application/json';
+        # 201 is created rather than ok
+        status 201;
+        return encode_json(
+            {
+                error => 0,
+                url   => '/settings/logo'
+            }
+        );
+    } else {
```
abeverley, line 572: Rather than this `else` statement, you could instead start the block with:
```
my $file = upload('file')
  or error __"No file provided";
```

lib/GADS/Schema/Result/Report.pm
```diff
@@ -335,4 +373,29 @@ sub fields_for_render {
...
+sub _read_security_marking {
+    my $self = shift;
+
+    my $marking = $self->security_marking;
+
+    return $marking if $marking;
+
+    my $instance = $self->instance;
+
+    return $instance->read_security_marking;
```
abeverley, line 398: The above could simply be:
```
return $self->security_marking || $self->instance->read_security_marking;
```

lib/GADS/Schema/Result/Report.pm
```diff
@@ -272,29 +286,34 @@ Function to create a PDF of the report - it will return a PDF object
...
+    my $pdf;
+
+    if($logo) {
+        $pdf    = CtrlO::PDF->new(
+            header => $marking,
+            footer => $marking,
+            logo   => $logo,
+        );
+        $pdf->add_page;
+        my $topmargin = -30;
```
abeverley, line 301: Does this require the logo to be a certain size? If so, it might be better retrieving the size of the logo and setting the margin accordingly?

lib/GADS/Schema/Result/Report.pm
```diff
@@ -272,29 +286,34 @@ Function to create a PDF of the report - it will return a PDF object
...
+        $pdf->heading( $self->title || $self->name, topmargin=>$topmargin );
+        $pdf->text($self->description , size => 14 ) if $self->description;
+    } else {
+        $pdf    = CtrlO::PDF->new(
+            header => $marking,
+            footer => $marking,
+        );
+        $pdf->add_page;
+        $pdf->heading( $self->title || $self->name );
+        $pdf->heading($self->description , size => 14 );
```
abeverley, line 311: The description heading in the previous block is only set if the description exists, where as it is set in this statement regardless? It might be worth combining these statements outside of the `if` statement to make them consistent? (and maybe others such as `add_page`)

lib/GADS/Schema/Result/Report.pm
```diff
@@ -304,12 +323,31 @@ sub create_pdf
...
     };
 
-    $pdf->table(
-        data         => $fields,
-        header_props => $hdr_props,
-        border_c     => '#007c88',
-        h_border_w   => 1,
-    );
+    foreach my $topic (keys %$grouped_data) {
+        next if $topic eq 'Other';
```
abeverley, line 327: What is the reason for excluding the "other" topic?

lib/GADS/Schema/Result/Report.pm
```diff
@@ -304,12 +323,31 @@ sub create_pdf
...
+        my $fields = [ [$topic, ''] ];
+        push( @{$fields}, [ $_->{name}, $_->{value} ] ) foreach (@{$grouped_data->{$topic}});
+
+        $pdf->table(
+            data         => $fields,
+            header_props => $hdr_props,
+            border_c     => '#007c88',
+            h_border_w   => 1,
+        );
+    }
```
abeverley, line 350: Sorry if I'm being stupid, the it looks like the same tables are printed twice?

lib/GADS/Schema/Result/Site.pm
```diff
@@ -331,4 +337,46 @@ sub user_fields
     return @fields;
 }
 
+sub load_logo {
+  my $self = shift;
+  my $filecheck = GADS::Filecheck->instance();
+
+  my $logo_path = $self->create_temp_logo;
+
+  return undef unless $logo_path;
```
abeverley, line 346: I would simply write: `my $logo_path = $self->create_temp_logo or return`;

lib/GADS/Schema/Result/Site.pm
```diff
@@ -331,4 +337,46 @@ sub user_fields
...
+
+  my $result = +{
+    'content_type'=> $filecheck->get_filetype($logo_path),
+    'filename' => 'logo',
+    'data' => $self->site_logo
+  };
```
abeverley, line 352: No need for `my $result`, just return the hashref directly.

lib/GADS/Schema/Result/Site.pm
```diff
@@ -331,4 +337,46 @@ sub user_fields
...
+}
+
+sub read_security_marking {
+  my $self = shift;
+  my $marking = $self->security_marking;
+
+  return $marking if $marking;
+
+  my $config = GADS::Config->instance;
+  $marking = $config && $config->gads && $config->gads->{header};
```
abeverley, line 377: See similar comments above, this section could probably be collapsed to a couple of lines.

views/fields/editable_text.tt
```diff
@@ -0,0 +1,28 @@
...
+        aria-describedby="update-button"
+        name = "[% name %]" />
+    <button
+        type="[% btn_type || 'button' %]"[% IF btn_val %]
+        value="[% btn_val %]"
+        id="[% btn_val %]"
+        name="[% btn_val %]"[% END %]
+        class="btn btn-success"
+        id="update-button">Update</button>
+</div>
```
abeverley, line 28: I'm wondering rather than creating a new field type, whether we can just use the same style as other settings pages, with the submit button at the bottom of the page?

abeverley (commented):
Thanks Dave, almost there! Comments inline.

lib/GADS/API.pm
droberts-ctrlo: As this is done using AJAX, I thought it more "sensible" to use the JSON as (it seemed to be) intended and include an error response to display to the user?

lib/GADS/Schema/Result/Report.pm
droberts-ctrlo: This is to print only the fields within the 'Other' topic (see other comment). I have updated this to make it clearer.

lib/GADS/Schema/Result/Report.pm
droberts-ctrlo: When this runs, for some reason, the placement of the 'Other' topic appears to be random each run - most runs it is rendered at the bottom, but for some it can be either the middle, or the top. I feel that said topic isn't as important as the named topics ('Other' denotes a field that has no topic) and as such should be rendered at the bottom.

lib/GADS/API.pm
abeverley: The above statement (using `error`) will automatically render JSON. See https://github.com/ctrlo/GADS/blob/master/lib/GADS/API.pm#L32 - so no need to manually write a JSON return statement :)

views/fields/editable_text.tt
droberts-ctrlo: This is fine, and doable - I created this when it was originally an AJAX request. The only issue I can percieve is that the report list table would then have what looks like a "hanging" submit button in the bottom right which would have an "unclear" purpose in my opinion, so it was clearer in my opinion (as it was only to change a single field) to do it this way. With that in mind for the report view, it was then added in the same manner to the settings view in order to preserve consistency.

lib/GADS/Schema/Result/Report.pm
droberts-ctrlo: Not in my opinion, because the margin-top is a relative mesurement, and I've made it so that the sub-heading is low enough that word-wrap around the logo is no longer a problem (it seemed to be the most efficient way to solve that problem).

abeverley: Okay, great.

lib/GADS/Schema/Result/Report.pm
```diff
@@ -304,12 +322,30 @@ sub create_pdf
...
-    $pdf->table(
-        data         => $fields,
-        header_props => $hdr_props,
-        border_c     => '#007c88',
-        h_border_w   => 1,
-    );
+    foreach my $topic (keys %$grouped_data) {
+        next if $topic eq 'Other';
+        my $fields = [ [$topic, ''] ];
+        push( @{$fields}, [ $_->{name}, $_->{value} ] ) foreach (@{$grouped_data->{$topic}});
```
abeverley, line 328: As discussed, please combine this code with the similar code in `GADS::Role::Presentation::Record`, moving the overall code in `GADS::Record`

lib/GADS.pm
```diff
@@ -1329,6 +1329,45 @@ any ['get', 'post'] => '/settings/team_edit/:id' => require_any_role [qw/useradm
...
+    my $site             = var 'site';
+
+    if(body_parameters->get('submit')) {
+        my $post_marking = body_parameters->get('security_marking');
+
+        $site->update({ security_marking => $post_marking });
+    }
+
+    my $logo             = $site->site_logo ? 1 : 0;
+    my $security_marking = $site->security_marking || config->{gads}->{header};
```
abeverley, line 1342: Should be able to use `$site->read_security_marking` instead and remove the `config`

views/fields/editable_text.tt
```diff
@@ -0,0 +1,28 @@
+[%-
```
abeverley, line 1: Is this file still needed?

abeverley (commented):
Almost there! Comments inline.

lib/GADS.pm
```diff
@@ -2807,6 +2846,13 @@ prefix '/:layout_name' => sub {
                 return forwardHome("$lo/report");
             }
 
+            if(body_parameters->get('submit')) {
+                my $security_marking = body_parameters->get('security_marking');
+                $layout->set_marking($security_marking);
```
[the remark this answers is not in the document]
droberts-ctrlo: Success not showing

lib/GADS/Schema/Result/Report.pm
```diff
@@ -304,12 +294,37 @@ sub create_pdf
...
+    my %include;
+
+    for my $val (@layouts) {
+        $include{$val->layout->name} = 1;
+    }
+
+    foreach my $col (@$d) {
+        next unless $include{$col->name};
+        push @$result, $col;
+    }
```
abeverley, line 309: I think this whole section could be simplified, something like (include comments too):
```
# Columns from the report
my %include = map { $_->layout_id => 1 } $self->report_layouts;
# Filter out columns the user does not have access to
my $result = [grep $include{$_->id}, @{$reord->columns_render}];
```

lib/GADS/Schema/Result/Report.pm
```diff
@@ -304,12 +294,37 @@ sub create_pdf
...
+
+    my @cols = $record->presentation_map_columns(columns=>$result);
+    my @topics = $record->get_topics(\@cols);
+
+    foreach my $topic (@topics) {
+        my $topic_value = defined $topic->{topic}? $topic->{topic} : undef;
+        my $topic_name = "Other";
+        $topic_name = $topic_value->name if $topic_value;
```
abeverley, line 320: These 3 lines could be written as:
```
my $topic_name = $topic->{topic} ? $topic->{topic}->name : 'Other';
```

PR #323, opened 2024-01-15
Bootstrap-select added

droberts-ctrlo:
Probably need to regenerate the lockfile, but mine is a mess, so I'll leave that to you! :smile:


abeverley:
Thanks Dave, I'm wondering whether this also needs something in `src/frontend/components/form-group/filter/lib/component.js` for the normal view filters?

PR #324, opened 2024-01-19
Mapper fixes

droberts-ctrlo:
Found that some trees in production have the same value multiple times. The `mapper` has now been updated to allow for this and filter out duplicate fields.

### DIFF
--- a/lib/GADS.pm
+++ b/lib/GADS.pm
@@ -4285,7 +4285,7 @@ prefix '/:layout_name' => sub {
 
         return to_json {
             error   => 0,
-            records => [ $column->values_beginning_with($query, noempty => query_parameters->get('noempty')) ],
+            records => [ $column->values_beginning_with($query, noempty => query_parameters->get('noempty')) ]
         }
     };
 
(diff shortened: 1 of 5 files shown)


abeverley:
As promised, I've been taking a look at this to see whether `values_beginning_with` could be folded into the existing functionality in `GADS::Column`. I'm actually wondering whether a really simple solution could be to change the existing function to always use a `max`. This should not affect existing column types, and should then allow the `group` to be added to the `resultset_for_values` in `GADS::Column::Tree`.

I.e. add the `select` from this PR into `GADS::Column` instead.

Can you try that please?

abeverley:
Thanks Dave, can you just remove the changes in `src` please before I merge? (I think they are just formatting changes?)

droberts-ctrlo:
They aren't, it's removal of the mapping JS that removed the duplicates that I removed at your request. :smile:

droberts-ctrlo:
Hi Andy, those are the changes for removal of duplicates that are no longer needed in the JS as it's now handled by the grouping in the backend. ((Apologies for previous comment, realised it may have come across as unintentionally blunt))

abeverley:
Great, thanks Dave.

PR #326, opened 2024-01-19
Added fullscreen control to audit table

droberts-ctrlo:
- Had to change table display DOM to display button
- Had to modify behaviour based on location.path as the button wouldn't re-enable without (even though code was being called correctly).


abeverley:
Thanks Dave, this looks good. However, rather than matching on the URL, can you see if you can switch on the fullscreen option with some other flag in the HTML please? Maybe a class or data attribute on the table? (I am not sure how the existing data table enables it - can that be replicated?)

PR #327, opened 2024-01-22
Fixed view for creating views

droberts-ctrlo:
- Added validation to dropdown for global views
- Removed admin view checkbox for users that aren't admin


abeverley:
I'd like to see if we can reuse the existing JS validation functions before merging this.

PR #333, opened 2024-01-26
(Possibly) fixed sidebar "jarring" load

droberts-ctrlo:
- Hide sidebar until load is complete, then show in the correct state
- Added transitions to ensure that pop out and in is "cleaner"

### DIFF
--- a/src/frontend/components/sidebar/_sidebar.scss
+++ b/src/frontend/components/sidebar/_sidebar.scss
@@ -2,6 +2,7 @@
   min-height: $min-height-container;
   padding: 50px 0 80px;
   color: $white;
+  transition: all 0.2s ease;
 
   &::before {
     content: '';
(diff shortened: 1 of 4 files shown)


abeverley:
Let's get the tests to pass before merging (requesting advice if need be)

tomhukins:
The test failures for Chromium and Firefox both describe the problem:
> `# Failed test 'The site navigation is visible after logging in'`
> `# at webdriver/t/login.t line 33.`

I see the same problem when running the tests in my development environment as described in `webdriver/README.md`.

When I visit the Web interface manually on this branch, I see no navigation at all.  This pull request needs to include a rebuild commit containing the changes generated by `yarn run build`.  After running this in my development environment, the tests pass.

So, on this occasion the tests have spotted the need to rebuild generated files.  I'm no fan of storing generated or compiled output in version control systems, but that's another story for another day.

PR #341, opened 2024-02-07
Fix for incorrect optimisation of Plotly JS

droberts-ctrlo:
Optimisation through Babel was causing globes to not work, Plotly has been removed from the optimisation process - tested and working locally


droberts-ctrlo:
The code for plotly needs to be excluded from Babel optimisation. This is to remain the same, do not remove this exclusion else the error `a.W is not a function` will throw again.

PR #344, opened 2024-02-08
Avoid using a module unnecesarily

tomhukins:
List::Util now supports many of the features List::MoreUtils used to provide.  List::Util ships as part of core perl, and is already likely to be loaded into the running process.

List::Util supports zip() from version 1.56, and uniq() from version 1.45.


tomhukins:
I pushed the first version of this too soon and it failed its tests.  This version works properly.

PR #347, opened 2024-02-21
Fix for rule filters not displaying on load correctly

droberts-ctrlo:


droberts-ctrlo:
This has been tested (by me) on a test system and appears to work
OT will test after merge and rebuild

PR #351, opened 2024-02-26
Updates to reporting as advised

droberts-ctrlo:
- Fixed display for curvals
- Added generation and last edited timestamps
- Open in new tab
- Secured URLs for invalid report instances
- Secured access to deleted reports
- Error added for incorrect file given for upload
- Logo now displays when present on load
- Created and Edited on added
- Added select all checkbox to fields
- Updated view to use simpler URL for historic reports.


abeverley:
Thanks Dave. Just one minor update - can you use `value` instead of `email` for saying who downloaded the report please?

PR #365, opened 2024-03-14
Calc fields now usable frontend

droberts-ctrlo:
Made calc fields usable regardless of permissions. These will now show on the DOM and will allow for calculations, but are visibly hidden from view - this is the only way I could perceive calcs working front-end whilst keeping them hidden.


droberts-ctrlo:
Going to review with a mind to fix JS testing

PR #367, opened 2024-03-14
Shrinking down site.js using chunks

droberts-ctrlo:
Have managed to get chunk for site.js down to 484KB from 763KB


.eslintignore
droberts-ctrlo, line 1: This is so I could ensure the code I was adding was "clean" - it is worth keeping in to ensure consistency with future code, although I'm open to correction.

.eslintrc.js
droberts-ctrlo, line 1: ESLint was nearly as old as Oliver - needed to update versions and rules in order to help keep (new) code up to standard

package.json
droberts-ctrlo, line 1: There were a _lot_ of deprecated and seeminly random "code-quality" tools here - in all honesty, it needed housekeeping, so as what I was doing in "sorting" chunks was pretty JS heavy, it seemed sensible to include the changes to this file to remove any irrelevant or deprecated guff.

src/frontend/components/button/lib/create-report-button.test.js
droberts-ctrlo, line 1: Tests updated to use "chunked" import mode in order to ensure that the component still works as expected.

src/frontend/components/calculator/index.js
droberts-ctrlo, line 1: As we have (previously) established that this component is rarely used (if at all), it seems more sensible to add it in chunked mode.

src/frontend/components/form-group/common/bootstrap-select.js
droberts-ctrlo, line 1: Disabling erroring of the below, else I'll be here all day fixing ESLint issues

src/frontend/components/form-group/select-widget/lib/component.js
```diff
@@ -104,16 +103,7 @@ class SelectWidgetComponent extends Component {
   }
 
   handleDocumentClick(e) {
-    const clickedOutside =
```
droberts-ctrlo, line 107: Not entirely sure this was me(?)

src/frontend/components/form-group/select-widget/lib/component.js
```diff
@@ -184,34 +173,38 @@ class SelectWidgetComponent extends Component {
     switch (key) {
       case 38: // UP
       case 40: // DOWN
-        const items = this.$available.find(".answer:not([hidden]) input")
-        let nextItem
+        {
```
droberts-ctrlo, line 176: Case statements now require (or possibly prefer) their own closure

src/frontend/components/modal/modals/new-table/lib/component.js
droberts-ctrlo, line 1: This file had a lot of "hanging" variables that were assigned but served no purpose - in the interest of minimising the code (as is the exercise here) these have been removed as required

src/frontend/js/lib/get-field-values.js
droberts-ctrlo, line 1: This branch is never hit as it's handled in a condition above

src/frontend/js/lib/util/filedrag/lib/filedrag.test.ts
droberts-ctrlo, line 1: Bit of cleanup while I'm here

src/frontend/js/lib/util/typeahead/lib/TypeaheadBuilder.test.ts
droberts-ctrlo, line 1: Again, cleanup

src/frontend/js/site.js
droberts-ctrlo, line 1: Removed summernote from here as it's loaded dynamically within the code.

tsconfig.json
droberts-ctrlo, line 1: Chunked import throws a fit in TS when JEST runs if I use a lower version - it's arbitrary, WebPack does the hard work for us!

yarn.lock
droberts-ctrlo, line 1: Updated version of TS and removed guff from the package file (and removed random dev requirements that were no longer required or used)

droberts-ctrlo (commented):
Futher to the comments included, I have also:
- Removed the unrequired usages of `self` that refer to `this` in order to shrink code (very) slightly and remove possibility of errors in the code
- Removed unnecessary variables where they are unused throughout the entire JS codebase

PR #368, opened 2024-03-15
Reporting edits

droberts-ctrlo:
- Removed header repeat[^1]
- Changed column width to fixed width for left hand side

[^1]: This is dependent on a `PDF::Table` code edit that is required due to a bug in their code

### DIFF
--- a/lib/GADS/Column.pm
+++ b/lib/GADS/Column.pm
@@ -30,7 +30,6 @@ use GADS::View;
 use HTML::Entities qw/encode_entities/;
 use MIME::Base64 qw/encode_base64/;
 use Text::Markdown qw/markdown/;
-use List::Util qw/uniq/;
 
 use Moo;
 use MooX::Types::MooseLike::Base qw/:all/;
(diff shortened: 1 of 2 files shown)


droberts-ctrlo:
We have discovered in testing that the placement of the "other" topic within the reports (and within the views) are dependent on the placement of the first field without a topic within the view. Therefore, placement is "up to the user".

In other words, should there be 3 fields in a instance, the first of those three denotes the placement of the "other" topic - if "field 1" has no topic, "other" will be first, if "field 2", then "other" will be second and so forth.

lib/GADS/Column.pm
```diff
@@ -1064,6 +1065,31 @@ sub delete
...
+    foreach my $reportLayout ($self->schema->resultset('ReportLayout')->search(
+            [
+                {layout_id => $self->id},
+            ]
+        )->all
+    ) {
+        #In theory, we would only reach here if there are layouts in a deleted report
+        $reportLayout->delete;
+    }
+
```
abeverley, line 1092: I think I would do the above something like this:
```
my $rs = $self->schema->resultset('ReportLayout')->search({
    layout_id => $self->id,
}, {
    prefetch => 'report',
});
my $rs_live = $rs->search({},{'report.deleted' => undef });
if ($rs_live->count)
{
    my %reports = map { $_->report_id => $_->report } $rs_live->all;
    my $r = join(q{, }, map {$_->name} values %reports);
    error __x"The following reports reference this field: \"{report}\". Please update them before deletion.",
        report => $r;
}
# We only reach here if there are layouts in deleted reports
$_->delete foreach $rs->all;
```

PR #369, opened 2024-03-21
Finished initial iteration of People filtering

droberts-ctrlo:
The reason it's "initial iteration" is I know (and am well aware) that some of it is "wrong" or could be better, but I'm lacking a direction. I hope through feedback, I can ascertain said direction.


lib/GADS/Helper/ConditionBuilder.pm
```diff
@@ -0,0 +1,105 @@
+use utf8;
+
+package GADS::Helper::ConditionBuilder;
```
abeverley, line 3: I like the use of this class in principle. However, I think we should also integrate it into `GADS::Records::_search_construct` so as to create common code.

lib/GADS/Helper/ConditionBuilder.pm
```diff
@@ -0,0 +1,105 @@
...
+        return $value;
+    }
+    elsif ( lc $field eq 'begins_with' ) {
+        return $value . '%';
+    }
+    elsif ( lc $field eq 'ends_with' ) {
+        return '%' . $value;
+    }
+    elsif ( lc $field eq 'contains' ) {
+        return '%' . $value . '%';
```
abeverley, line 46: As `GADS::Records::_search_construct` shows, `$value` will need to be escaped

lib/GADS/Users.pm
```diff
@@ -101,7 +107,9 @@ sub user_summary_rs
 
 sub _build_all
 {   my $self = shift;
-    my @users = $self->user_summary_rs->all;
+    my @users;
+    @users = $self->user_summary_rs->all unless $self->person_filter;
+    @users = $self->user_summary_rs->search($self->person_filter)->all if $self->person_filter;
```
abeverley, line 112: This could be better written as:
```
my $users = $self->user_summary_rs;
$users = $users->search($self->person_filter) if $self->person_filter;
[$users->all];
```

lib/GADS/PeopleFilter.pm
```diff
@@ -0,0 +1,43 @@
...
+use warnings;
+
+use Moo;
+use Log::Report 'linkspace';
+use Encode;
+use MIME::Base64;
+
+extends 'GADS::Filter';
+
+sub base64 {
```
abeverley, line 15: I'm not sure that this whole function is needed and/or the one in the parent class could be used/combined?

abeverley (commented):
Thanks - in principle looking like a good approach. Just a bit of work needed to try and reduce code duplication.

lib/GADS/Column/Person.pm
```diff
@@ -195,5 +211,44 @@ sub import_value
...
+      :                   {};
+    if ($resultset) {
+        my $filter       = $self->filter;
+        my $match_result = $resultset->search(
+            $search,
+            {
+                rows => 10,
+            },
+        );
+        @value = map {
```
droberts-ctrlo, line 239: This is where this implementation differs (aside from the inclusion of the filter above) - as this doesn't use grouping in the same manner as the original it is felt that this should remain here, rather than bastardising the original implementation

lib/GADS/Filter.pm
droberts-ctrlo, line 1: For some reason, the filter in general returns the correct values in `$self->as_hash`, but the `$self->base64` returns the values with no rules included.

lib/GADS/Column.pm
```diff
@@ -1546,28 +1546,40 @@ sub validate_search_date
 # possible (e.g. integer)
 sub resultset_for_values {};
 
-sub values_beginning_with
-{   my ($self, $match_string, %options) = @_;
+sub match_result {
```
droberts-ctrlo, line 1549: This is a sub encapsulating the common part of `values_beginning_with` between Column and Person

lib/GADS/Helper/ConditionBuilder.pm
droberts-ctrlo: This is to be implemented

lib/GADS/Column/Person.pm
```diff
@@ -23,12 +23,30 @@ use GADS::Users;
 use Moo;
 use MooX::Types::MooseLike::Base qw/:all/;
 
+use Data::Dump qq/pp/;
```
abeverley, line 26: Remove dump

lib/GADS/Filter.pm
```diff
@@ -120,24 +124,25 @@ sub _clear_lazy
...
-        if ($col->has_filter_typeahead)
-        {
-            $filter->{data} = {
-                text => $col->filter_value_to_text($filter->{value}),
-            };
-        }
-        if ($col->type eq 'filval')
-        {
-            $filter->{filtered} = $col->related_field_id,
+    if(@{$self->filters}) {
```
abeverley, line 132: Is this condition necessary? The loop before would not be run anyway in such circumstances

lib/GADS/Helper/ConditionBuilder.pm
abeverley: I will take a look at this

lib/GADS/Users.pm
```diff
@@ -24,10 +24,18 @@ use Log::Report 'linkspace';
 use POSIX ();
 use Scope::Guard qw(guard);
 use Text::CSV;
+use GADS::Helper::ConditionBuilder;
 
 use Moo;
 use MooX::Types::MooseLike::Base qw(:all);
 
+use Data::Dump qq\pp\;
```
abeverley, line 32: Remove this

lib/GADS/Filter.pm
droberts-ctrlo: I believe this was for a slightly different implementation I previously tried that didn't work - now removed. :smile:

PR #377, opened 2024-04-03
Updated Readme.md and added Rexfile

droberts-ctrlo:
Readme has comments and asides that will (obvs.) need sorting - also wording is a bit pants, all requires review before publishing or use.


droberts-ctrlo:
Possibly will need some proof-reading but should be good to go. :smile:

abeverley:
Looks great - can you just change UIUX branch to dev please?

droberts-ctrlo:
> Looks great - can you just change UIUX branch to dev please?

It is

PR #385, opened 2024-04-15
Added purge classes and script

droberts-ctrlo:


droberts-ctrlo:
Closing in favour of #386

PR #386, opened 2024-04-16
Record purging complete

droberts-ctrlo:
It is felt that this is possibly not the "preferred" way to perform the task, but I felt that this was the best way forward (and the most efficient).


droberts-ctrlo:
There is quite a bit of "refinement" to the `Purge` packages here - it is up to the maintainer if #385 is felt to be still required or if that is to be bypassed in favour of this.

lib/GADS.pm
```diff
@@ -3024,6 +3025,99 @@ prefix '/:layout_name' => sub {
...
+            include_children    => 1,
+            view_limit_extra_id => undef, # Override any value that may be set
+            view                => $view,
+        );
+
+        my $params = +{};
+        my $record_data = [];
+
+        if(param 'submit') {
+            my @columns = body_parameters->get_all('record');
```
abeverley, line 3048: I think `column_id` would be a better name than `record`, otherwise this is confusing.

lib/GADS.pm
```diff
@@ -3024,6 +3025,99 @@ prefix '/:layout_name' => sub {
...
+            $records->columns([@columns]);
+        } elsif(param 'purge') {
+            my @columns = split(/,/,body_parameters->get('records'));
```
abeverley, line 3051: This would be better as multiple hidden fields, and then it could be combined with the block above.

lib/GADS.pm
```diff
@@ -3024,6 +3025,99 @@ prefix '/:layout_name' => sub {
...
+            id => 1,
+            createddate => 1,
+            createdby => 1,
+            serial => 1,
+        );
+
+        unless(param 'purge') {
+            while (my $record = $records->single) {
+                foreach my $column (@{$record->presentation->{columns}}) {
+                    next if $ignores{$column->{type}};
```
abeverley, line 3065: There should be a way of using a column attribute instead, such as `internal`

lib/GADS.pm
```diff
@@ -3024,6 +3025,99 @@ prefix '/:layout_name' => sub {
...
+            my @columns = split(/,/,body_parameters->get('records'));
+            $records->columns([@columns]);
+        }
+
+        my %ignores = (
+            id => 1,
+            createddate => 1,
+            createdby => 1,
+            serial => 1,
+        );
```
abeverley, line 3060: See comment below

lib/GADS.pm
```diff
@@ -3024,6 +3025,99 @@ prefix '/:layout_name' => sub {
...
+        unless(param 'purge') {
+            while (my $record = $records->single) {
+                foreach my $column (@{$record->presentation->{columns}}) {
+                    next if $ignores{$column->{type}};
+                    my $result = +{};
+                    $result->{col_id} = $column->{id};
+                    $result->{name} = $column->{name};
+                    $result->{id} = $column->{data}->{record_id};
+                    $result->{value} = $column->{data}->{value};
+                    push @$record_data, $result;
```
abeverley, line 3071: Styling: I think this would be better written as:
```
push @record_data, {
    col_id => $column->{id},
    ....
};
```
It would also be tidier to use an array and then only turn it into a reference when needed.

lib/GADS.pm
```diff
@@ -3024,6 +3025,99 @@ prefix '/:layout_name' => sub {
...
+                }
+            }
+        }
+
+        $params->{page} = 'initial_purge';
```
abeverley, line 3076: Move this to the `else` block

lib/GADS.pm
```diff
@@ -3024,6 +3025,99 @@ prefix '/:layout_name' => sub {
...
+
+        if(param 'submit') {
+            $params->{page} = 'purge_stage_2';
+            $params->{records} = $record_data;
+            my @mapped_ids = map {$_->{col_id}} @$record_data;
+            my @mapped_records = uniq map {$_->{id}} @$record_data;
+            $params->{column_list} = join(',', @mapped_ids);
+            $params->{record_list} = join(',', @mapped_records);
```
abeverley, line 3084: I'm not sure the above is needed - can this be done on the fly in the template? As per previous comment, the parameters would be better rendered as multiple hidden fields.

lib/GADS.pm
```diff
@@ -3024,6 +3025,99 @@ prefix '/:layout_name' => sub {
...
+        } elsif(param 'purge') {
+            my $columns = body_parameters->get('columns') or error __"Unable to get columns";
+            my $rids = body_parameters->get('records') or error __"Unable to get records";
+            my @mapped_ids = split(/,/,$columns);
+            my @records = split(/,/,$rids);
```
abeverley, line 3089: I'd prefer to use the records from the view (simpler), unless the idea is to prevent records being purged that were not in the view originally?

lib/GADS.pm
```diff
@@ -3024,6 +3025,99 @@ prefix '/:layout_name' => sub {
...
+                        next unless $d->{name} eq $val->{name};
+                        $d->{count}++;
+                    }
+                }
+                else {
+                    $seen->{$val->{name}} = 1;
+                    push @$result, {id=>$val->{col_id}, name => $val->{name}, count => 1};
+                }
+            }
+            $params->{columns} = $result;
```
abeverley, line 3115: I think there should be a way of removing most of the above and doing a simpler version in the template?

lib/GADS/Role/Purgable.pm
```diff
@@ -0,0 +1,31 @@
...
+
+sub purge {
+    my $self = shift;
+    
+    my $source = $self->recordsource or error __"No recordsource defined";
+    my @field = $self->valuefield or error __"No valuefield defined";
+    my $schema = $self->result_source->schema;
+
+    $schema->txn_do(sub {
+        my $rs = $schema->resultset($source)->find($self->id);
```
abeverley, line 26: I'm not sure the above is necessary (along with the definition of `recordsource`). You should simply be able to do `$self->update(...)`?

lib/GADS/Schema/Result/File.pm
```diff
@@ -183,6 +189,26 @@ sub export_hash
 # Created by DBIx::Class::Schema::Loader v0.07043 @ 2015-11-13 16:02:57
 # DO NOT MODIFY THIS OR ANYTHING ABOVE! md5sum:266q8eJmPiCjcNhwddaUkw
 
+sub _build_recordsource { 'File'; }
+sub _build_valuefield { undef; }
+
+# I've done this to override the role - I don't know how else to best implement this!
```
abeverley, line 195: I know we discussed this, but I can't remember the reason for putting "purged" in a file but blanking all other fields?

abeverley (commented):
Thanks Dave, this is pretty good on the whole. I think there are just a few ways of simplifying things.

lib/GADS/Schema/Result/File.pm
droberts-ctrlo: In order to show on cronology that a file was uploaded and when (sometimes the file is uploaded later than the record being created)

lib/GADS.pm
```diff
@@ -3024,6 +3025,89 @@ prefix '/:layout_name' => sub {
...
+            my @columns = body_parameters->get_all('column_id');
+            $records->columns(\@columns);
+        }
+
+        unless(param 'purge') {
+            while (my $record = $records->single) {
+                foreach my $column (@{$record->columns_render}) {
+                    next if $column->internal;
+
+                    my $clone = $record->clone;
```
droberts-ctrlo, line 3061: This is to count the "correct" number of versions, rather than using `$record->versions`

lib/GADS/Column/String.pm
```diff
@@ -26,6 +26,11 @@ extends 'GADS::Column';
 
 with 'GADS::Role::Presentation::Column::String';
 
+sub purge {
```
droberts-ctrlo, line 29: I'm not entirely sure what this is for - it may be an old debug value that was placed in before we moved to using roles - I shall take a look.

src/frontend/components/data-table/lib/component.js
```diff
@@ -588,6 +588,7 @@ class DataTableComponent extends Component {
   getConf() {
     const confData = this.el.data('config')
     let conf = {}
+    const self = this;
```
droberts-ctrlo, line 591: I didn't intend to push this - I shall remove this change as it _should_ have been resolved in a different PR

views/test.lua
droberts-ctrlo, line 1: Shall remove - I don't quite know what this is from, but (obviously) it is not required here.

lib/GADS.pm
```diff
@@ -3024,6 +3025,89 @@ prefix '/:layout_name' => sub {
...
+                }
+            }
+        }
+
+        if(param 'submit') {
+            $params->{records} = \@record_data;
+            my @mapped_ids = map {$_->{col_id}} @record_data;
+            my @mapped_records = uniq map {$_->{id}} @record_data;
+            $params->{column_list} = \@mapped_ids;
+            $params->{record_list} = \@mapped_records;
```
abeverley, line 3086: I haven't looked at this in detail, but I think there could be some improvements here, to prevent having one parameter built from another parameter. I think probably better to have separate variables that are populated as required. E.g. push the record IDs onto a separate array in the $record loop rather than the $column loop. Similar for the column IDs, I think this could be `$params->{column_list} = [map $_->id, $records->columns_render]` or similar.

lib/GADS.pm
abeverley: Unfortunately I think use `clone()` and `find_chronology_id` are going to make this very slow for large systems and therefore not an option. The latter is already very slow for single records, so doing it across many records will be problematic I think.

lib/GADS.pm
```diff
@@ -3024,6 +3025,89 @@ prefix '/:layout_name' => sub {
...
+            return template 'historic_purge/confirm' => $params;
+        } elsif(param 'purge') {
+            my @columns = body_parameters->get_all('columns') or error __"Unable to get columns";
+            my @rids = body_parameters->get_all('column_id') or error __"Unable to get records";
```
abeverley, line 3090: Not sure why this is `column_id` when it talks about records?

lib/GADS.pm
```diff
@@ -3024,6 +3025,89 @@ prefix '/:layout_name' => sub {
...
+            my $total = 0;
+            foreach my $r (@rids) {
+                my $record = schema->resultset('Record')->find($r);
+                my $current_id = $record->current_id;
+                my $current = schema->resultset('Current')->find($current_id);
+                $current->historic_purge(@columns);
```
abeverley, line 3096: This doesn't look quite right. It should be possible to obtain the current_id directly?

lib/GADS/Role/Purgable.pm
```diff
@@ -0,0 +1,29 @@
...
+
+use strict;
+use warnings;
+
+use Moo::Role;
+
+has recordsource => (
+    is      => 'lazy',
+    builder => sub { undef; }
+);
```
abeverley, line 11: There should be no need for `recordsource` anymore. The only place it is used (`GADS::Schema::Result::File`) can obtain it from itself.

lib/GADS/Schema/Result/File.pm
```diff
@@ -183,6 +189,26 @@ sub export_hash
...
+sub purge {
+    my $self = shift;
+
+    my $source = $self->recordsource or error __"No recordsource defined";
+    my $schema = $self->result_source->schema;
+
+    $schema->txn_do(sub {
+      my $file = $schema->resultset($source)->search({},{prefetch=>'value'})->find($self->id);
+      my $value = $file->value;
+      $value->update({
```
abeverley, line 205: I think this could simply be `$self->update` and remove the previous 2 lines?

abeverley (commented):
Thanks Dave, I think we are almost there with this. There may be a couple of other minor things, but I've commented on the main things that need considering.

lib/GADS.pm
droberts-ctrlo: I'm inclined to agree, in retrospect!

lib/GADS/Schema/Result/File.pm
droberts-ctrlo: I feel I should add context to why I do this - other languages I've previously worked with don't allow this, so it's an old habit, I shall try to remember

lib/GADS.pm
droberts-ctrlo: Looking at this a bit deeper, I am wondering if I could do this in the view, rather than here using the `@record_data` (as that is already passed to the view) rather than passing the same thing 3 different ways - I'll take a look at both methods and work out which is best.

lib/GADS.pm
```diff
@@ -3024,6 +3024,62 @@ prefix '/:layout_name' => sub {
...
+            my $ids = [];
+
+            while (my $record = $records->single)
+            {
+                my @mapped_columns =$record->presentation_map_columns(columns => \@columns);
+
+                foreach my $column (@mapped_columns)
+                {
+                    push @$ids, $column->{id};
+                    $table_data->{ $record->current_id }->{ $column->{name} } = $column->{data}->{value} || "No value";
```
abeverley, line 3060: Please use field ID rather than field name, in case of duplicates. The column name can then be retrieved either from the layout object or probably better by adding to `$table_data`.

lib/GADS/Role/Purgable.pm
```diff
@@ -0,0 +1,29 @@
+package GADS::Role::Purgable;
+
+use strict;
+use warnings;
+
+use Moo::Role;
+
+has valuefield => (
+    is      => 'lazy',
+    builder => sub { ('value'); }
```
abeverley, line 10: I'm not sure that arrays work reliably in Moo, I think better to use an arrayref, and then also set `isa` to `ArrayRef`

lib/GADS/Role/Purgable.pm
```diff
@@ -0,0 +1,29 @@
...
+);
+
+sub purge {
+    my ($self,$user) = @_;
+    
+    my @field = $self->valuefield or error __"No valuefield defined";
```
abeverley, line 16: Better to use a `panic` here as it should never happen

lib/GADS/Role/Purgable.pm
```diff
@@ -0,0 +1,29 @@
+package GADS::Role::Purgable;
+
+use strict;
+use warnings;
+
+use Moo::Role;
+
+has valuefield => (
```
abeverley, line 8: A better name would be `value_fields`

lib/GADS/Schema/Result/Date.pm
```diff
@@ -13,7 +13,10 @@ GADS::Schema::Result::Date
 use strict;
 use warnings;
 
-use base 'DBIx::Class::Core';
+use Moo;
+
+extends 'DBIx::Class::Core';
+sub BUILDARGS { $_[2] || {} }
```
abeverley, line 19: This class seems to be missing `purged_by` and `purged_on`

lib/GADS/Schema/Result/File.pm
```diff
@@ -183,6 +219,23 @@ sub export_hash
 # Created by DBIx::Class::Schema::Loader v0.07043 @ 2015-11-13 16:02:57
 # DO NOT MODIFY THIS OR ANYTHING ABOVE! md5sum:266q8eJmPiCjcNhwddaUkw
 
+sub _build_valuefield { undef; }
+
+# I've done this to override the role - I don't know how else to best implement this!
```
abeverley, line 224: I think better to use Moo's `after`, especially as the `value` will still need nulling anyway?

lib/GADS/Schema/Result/File.pm
```diff
@@ -183,6 +219,23 @@ sub export_hash
...
+    my $self = shift;
+
+    my $source = $self->recordsource or error __"No recordsource defined";
+    my $schema = $self->result_source->schema;
+
+    $schema->txn_do(sub {
+      $self->update({
+          name=>'purged',
+          mimetype=>'text/plain',
+          content=>encode('utf-8', 'purged'),
```
abeverley, line 235: Formatting :)

lib/GADS/Schema/Result/Person.pm
```diff
@@ -179,6 +213,7 @@ sub export_hash
     };
 }
 
+sub _build_recordsource { 'Person'; }
```
abeverley, line 216: Is this still needed?

lib/GADS/Schema/Result/Ragval.pm
```diff
@@ -141,5 +175,7 @@ sub sqlt_deploy_hook {
     $sqlt_table->add_index(name => 'ragval_idx_value', fields => ['value']);
 }
 
+sub _build_recordsource { 'Ragval'; }
```
abeverley, line 178: Still needed?

lib/GADS/Schema/Result/String.pm
```diff
@@ -144,4 +178,6 @@ sub export_hash
     };
 }
 
+sub _build_recordsource { 'String'; }
```
abeverley, line 181: Still needed?

lib/GADS/Schema/ResultSet/Current.pm
```diff
@@ -15,6 +15,18 @@ __PACKAGE__->load_components(qw/
     /
 );
 
+sub historic_purge {
+    my ($self, $user, $current_ids, $layouts) = @_;
+
+    error __"Please select some values to delete" if !$current_ids || !@$current_ids;
+    error __"Please select some layouts to delete" if !$layouts || !@$layouts;
+
+    my @result = $self->search({id=>{-in=>$current_ids}})->all;
```
abeverley, line 24: This can simply be: `$self->search({ id => $current_ids })`

src/frontend/components/data-table/lib/component.js
```diff
@@ -1,252 +1,252 @@
 /* eslint-disable @typescript-eslint/no-this-alias */
 
-import { Component, initializeRegisteredComponents } from 'component'
```
abeverley, line 3: Can we keep the linting out of this PR please?

abeverley (commented):
Almost there! Also, I have just realised that we need to include at least some basic backend testing.

lib/GADS/Role/Purgable.pm
```diff
@@ -0,0 +1,30 @@
...
+);
+
+sub purge {
+    my ($self,$user) = @_;
+    
+    my @field = @{$self->value_fields} or panic __"No valuefield defined";
+    my $schema = $self->result_source->schema;
+
+    my $purge_needed = grep { $self->$_ } @field;
+    $purge_needed = 1 if ref($self) eq "Gads::Schema::Result::File";
```
droberts-ctrlo, line 23: Don't believe this is needed now it's using "around" but not certain so leaving it in for now.

abeverley: Instead of this, I would create a new method called `is_purged` and override it in `GADS::Datum::File`

lib/GADS/Schema/Result/Enum.pm
```diff
@@ -179,5 +213,7 @@ sub export_hash
     };
 }
 
+sub _build_valuefield { ('value'); }
```
abeverley, line 216: Is this still needed?

lib/GADS/Datum/File.pm
```diff
@@ -153,10 +153,27 @@ sub _build_files
...
 
     return \@return;
 }
 
+has _is_purged => (
+    is=>'lazy',
+    builder => sub {
+        my $self = shift;
+        my $id = $self->record_id;
+        my $datum = $self->schema->resultset('File')->find($id) if $id;
```
abeverley, line 171: I don't think this will return the correct datum. I think you will need a new method which returns the applicable `GADS::Schema::Result::File`, e.g.
```
sub _datums
{   my $self = shift;
    $self->schema->resultset('File')->search({
        record_id => $self->record_id,
        layout_id => $self->column->id,
    })->all;
}
```

Then create a new `is_purged` method

lib/GADS/Schema/Result/File.pm
```diff
@@ -183,6 +219,26 @@ sub export_hash
...
+
+    print STDERR "Before purge...\n";
+
+    my $field = $self->value_fields->[0];
+
+    print STDERR "Get field: $field\n";
+
+    my $val = $self->$field;
+
+    print STDERR "Get value: $val\n";
```
abeverley, line 234: Remove debug

lib/GADS/Schema/Result/Intgr.pm
```diff
@@ -141,5 +175,7 @@ sub export_hash
     };
 }
 
+sub _build_recordsource { 'Intgr'; }
```
abeverley, line 178: Still needed?

lib/GADS.pm
```diff
@@ -3024,6 +3028,62 @@ prefix '/:layout_name' => sub {
...
+            my $ids = [];
+
+            while (my $record = $records->single)
+            {
+                my @mapped_columns =$record->presentation_map_columns(columns => \@columns);
+
+                foreach my $column (@mapped_columns)
+                {
+                    push @$ids, $column->{id};
+                    $table_data->{ $record->current_id }->{ $column->{id} }->{ $column->{name} } = $column->{data}->{value} || "No value";
```
abeverley, line 3064: I think I would prefer `$table_data` to be an array not a hash. That saves the use of `keys` in the template and also retains the ordering

lib/GADS/Schema/Result/File.pm
```diff
@@ -183,6 +219,26 @@ sub export_hash
...
+    $self->result_source->schema->txn_do(
+        sub {
+            $orig->( $self, @_ );
+            $val->delete if $val;
+        }
+    );
+};
```
droberts-ctrlo, line 241: I Hate files!

src/frontend/components/data-table/_data-table.scss
droberts-ctrlo, line 1: I have _no_ idea why this is in here!

src/frontend/components/data-table/lib/component.js
droberts-ctrlo, line 1: Or this

lib/GADS.pm
```diff
@@ -1703,6 +1703,8 @@ any ['get', 'post'] => '/file/:id?' => require_login sub {
     # ID will either be in the route URL or as a delete parameter
     my $id = route_parameters->get('id') || body_parameters->get('delete')
         or error "File ID missing";
+    
+    return send_file( "Purged File", content_type => "text/plain", filename => "purged" ) if $id == -1;
```
abeverley, line 1707: Please test, I think this still needs to be referenced using a backslash

lib/GADS.pm
```diff
@@ -3028,6 +3030,69 @@ prefix '/:layout_name' => sub {
...
+
+            while (my $record = $records->single)
+            {
+                my @mapped_columns =$record->presentation_map_columns(columns => \@columns);
+                
+                my $values = {};
+
+                foreach my $column (@mapped_columns)
+                {
+                    $columns->{$column->{id}} = $column->{name};
```
abeverley, line 3067: Rather than populating `$columns` here (repeatedly), I think you should be able to move it out of the loop and use `@columns` instead?

lib/GADS.pm
```diff
@@ -3028,6 +3030,69 @@ prefix '/:layout_name' => sub {
...
+            my $layouts     = [body_parameters->get_all('layout')];
+
+            if ( process( sub { schema->resultset('Current')->historic_purge($user, $current_ids, $layouts) } ) )
+            {
+                return forwardHome({ success => "Records have now been purged" }, $layout->identifier . '/data');
+            }
+        }
+
+        my $records = GADS::Records->new(%params);
+        my @columns = grep { !$_->internal } @{ $records->columns_render };
```
abeverley, line 3091: This is repeated above in the first condition block, maybe move this one to the top and reuse?

lib/GADS/Datum/File.pm
```diff
@@ -153,10 +153,30 @@ sub _build_files
     elsif ($self->has_ids) {
         @return = $self->_ids_to_files(@{$self->ids});
     }
+    
+    return [+{
+        id => -1,
+        name => 'Purged',
+        mimetype => 'text/plain',
+    }] if $self->is_purged;
```
abeverley, line 161: Move this above the previous block, otherwise the previous block will be processed but do nothing

lib/GADS/Datum/File.pm
```diff
@@ -153,10 +153,30 @@ sub _build_files
...
+    [$self->schema->resultset('File')->search({
+        record_id => $self->record_id,
+        layout_id => $self->column->id,
+    })->all];
+}
+
+sub is_purged {
+    my $self = shift;
+    my @datums = @{$self->_datums};
+    return grep { $_->purged_on } @datums;
```
abeverley, line 177: This should use `is_purged` in order to keep things consistent with other classes

abeverley (commented):
Thanks Dave, great work, almost there!!

lib/GADS/Datum/File.pm
droberts-ctrlo: This is because files are handled differently, or am I having a "moment"?

lib/GADS.pm
```diff
@@ -3028,6 +3030,64 @@ prefix '/:layout_name' => sub {
...
+            schema => schema,
+            view   => $view,
+        );
+
+        my $records = GADS::Records->new(%params);
+        my @columns = grep { !$_->internal } @{ $records->columns_render };
+
+        if (defined param('stage1'))
+        {
+            $params{columns} = [ body_parameters->get_all('column_id') ];
```
abeverley, line 3054: The above line doesn't seem to do anything? Should it be passed to `$records`?

lib/GADS/Datum/File.pm
```diff
@@ -157,6 +163,20 @@ sub _build_files
     return \@return;
 }
 
+sub _datums
```
abeverley, line 166: Sorry, can we call this something else please? Maybe `_db_files` or something?

lib/GADS.pm
```diff
@@ -3028,6 +3030,71 @@ prefix '/:layout_name' => sub {
...
+        );
+
+        my $records = GADS::Records->new(%params);
+        my @columns = grep { !$_->internal } @{ $records->columns_render };
+        my $columns_selected = +{};
+
+        if (defined param('stage1'))
+        {
+            $records->columns([ body_parameters->get_all('column_id') ]);
+
```
droberts-ctrlo, line 3056: I don't know why I didn't think of this way before.

views/fields/sub/checkbox.tt
```diff
@@ -11,6 +11,7 @@
     [%- IF is_disabled -%]disabled[%- END -%]
     aria-labelledby="[% id %]-label"
     data-value="[% INCLUDE fields/sub/filter.tt value=label; %]"
+    [%- IF required -%]required="required"[%- END -%]
```
abeverley, line 14: Can you change this to `is_required` for consistency please?

lib/GADS.pm
```diff
@@ -3028,6 +3030,78 @@ prefix '/:layout_name' => sub {
...
+            view   => $view,
+        );
+
+        my $records = GADS::Records->new(%params);
+        my @columns = grep { !$_->internal } @{ $records->columns_render };
+        my $columns_selected = +{};
+
+        if (defined param('stage1'))
+        {
+            my $cols = {};
```
abeverley, line 3055: I think use `$columns_selected` here instead?

t/026_purge.t
```diff
@@ -0,0 +1,139 @@
...
+    foreach my $rec_rs ($current->records)
+    {
+        my $datum = $schema->resultset($col->{type})->search({
+            record_id => $rec_rs->id,
+            layout_id => $string_col->id,
+        })->next;
+        ok(!$datum->value, "Datum is now blank");
+        ok($datum->purged_by, "Purged by is set");
+        ok($datum->purged_on, "Purged on is set");
+        ok($datum->is_purged, "Is purged is set");
```
abeverley, line 135: Sorry, final final thing. Can you add a test to also check for values that should not have been purged? That is just as important as making sure that values have been purged :) Maybe something like:
```
foreach my $datum ($schema->resultset($col->{type})->search({
    record_id => $rec_rs->id,
    layout_id => { '!='  => $string_col->id },
})->all
{
    ok($datum->value, "Datum is not blank");
    ok(!$datum->purged_by, "Purged by is not set");
    ok(!$datum->purged_on, "Purged on is not set");
    ok(!$datum->is_purged, "Is purged is not set");
}
```

t/026_purge.t
```diff
@@ -0,0 +1,139 @@
...
+            mimetype => 'text/plain',
+            content  => 'Text file2',
+        },
+        as_string => "file2.txt",
+    },
+);
+
+foreach my $col (@cols)
+{
+    my $string_col = $columns->{ $col->{name} };
```
abeverley, line 88: And please change the name of this as it's confusing :)

t/026_purge.t
```diff
@@ -0,0 +1,139 @@
...
+    {
+        my $datum = $schema->resultset($col->{type})->search({
+            record_id => $rec_rs->id,
+            layout_id => $string_col->id,
+        })->next;
+        ok(!$datum->value, "Datum is now blank");
+        ok($datum->purged_by, "Purged by is set");
+        ok($datum->purged_on, "Purged on is set");
+        ok($datum->is_purged, "Is purged is set");
+    }
```
droberts-ctrlo, line 136: I genuinely cannot get this to allow me to create new values; I have tried adding `string2=>'keep';` above (at the top in `data`) and querying that, and get told the column doesn't exist, using trees (won't let me write or create them), using random data, and even sacrificing a goat (I cleaned and disinfected the table after)

t/026_purge.t
droberts-ctrlo: I genuinely cannot get this to allow me to create new values; I have tried adding string2=>'keep'; above (at the top in data) and querying that, and get told the column doesn't exist, using trees (won't let me write or create them), using random data, and even sacrificing a goat (I cleaned and disinfected the table after)

abeverley: You can leave the poor goat alone :) You'll need to specify 2 columns for the string, so something like the below. The tree should also work as follows:
```
my $sheet = Test::GADS::DataSheet->new(
    column_count => {
        string => 2,
    },
    data => [ {
        string1      => 'Foo',
        string2      => 'Keep',
        integer1     => 50,
        date1        => '2014-10-10',
        enum1        => 1,
        tree1        => 'tree1',
        daterange1   => [ '2012-02-10', '2013-06-15' ],
        person1      => 1,
        file1        => {
            name     => 'file1.txt',
            mimetype => 'text/plain',
            content  => 'Text file1',
        },
    } ],
);
```

abeverley:
Thanks Dave. Final thing - did you try my suggestion for adding the tree values?

droberts-ctrlo:
There is an error in this PR (typo and use of variable before declaration) - I'll do a different PR for this.

Cypress and WebDriver both picked up on this, I didn't :disappointed:

PR #392, opened 2024-04-23
Calc dropdown error fix

droberts-ctrlo:
Made error for client-side and moved button functionality into new file


abeverley:
In addition to this error handling, I think it also worth making the "not shown" option be the default selected drop-down option for a calc field.

PR #394, opened 2024-04-25
Updated Cypress version

droberts-ctrlo:
Cypress version update appears to fix FireFox issue - will continue to monitor, and should this become a "regular" issue, then a different framework (with compatible logic) may be necessary rather than persistent updating of libraries as an extraneous activity.


droberts-ctrlo:
Found that this version does _not_ fix the issue - closing PR as not felt this is required until a reliable fix has been implemented

PR #401, opened 2024-04-30
Fix for button component not saving correctly

droberts-ctrlo:
- Also included a fix for the datatable component saying "has" is not defined


droberts-ctrlo:
This fix was merged as part of PR#402

PR #412, opened 2024-05-21
Fixed bug in arrow display

droberts-ctrlo:
z-index was incorrect on the popover arrow, causing it to display with a border incorrectly, this has now been rectified


abeverley:
Thanks Dave. I'm just wondering whether the z-index value should match one of the existing z-indexes for the related popover?

abeverley:
Merged separately 13f1b582

PR #416, opened 2024-06-05
Contextual help added to view creation

droberts-ctrlo:
Visual issues due to upgrade of DataTables to version 2+


abeverley:
Thanks for this Dave. I think the only thing I'd like to see changed, is rather than having a generic `data` parameter, is to instead have a specific help parameter. E.g.
```
help = {
    "text" => "The name of the view is used to identify it in the list of views."
    "target" => "help__card"
    "title" => "Name of the view"
}
```

abeverley:
Sorry Dave, can you individually specify each help parameter please (rather than loop through the keys). I think it's just good (in general) to have fixed parameters for the templates.

PR #430, opened 2024-07-04
Moved to named savepoints for record locking

droberts-ctrlo:


lib/GADS/Datum/Code.pm
```diff
@@ -131,14 +131,13 @@ sub _write_unique
...
-                $schema->storage->svp_release;
+            $schema->storage->svp_rollback("sp_uq_calc");
+            $schema->storage->svp_release("sp_uq_calc");
         }
         elsif ($@) {
             $@->reportAll;
-        }
-        else {
-            $schema->storage->svp_release;
+        } else{
```
abeverley, line 139: Thanks Dave, looks good, please just revert the formatting of the `else` :)

PR #435, opened 2024-07-11
Report permissions complete

droberts-ctrlo:
- Schema version will clash with purging - whichever is merged second of the two will need to have it's version incremented as appropriate and new migration generated
- Perltidy has been run on the report class and other classes where it has been deemed appropriate (i.e. changes to formatting have minimal impact)


lib/GADS/Schema/Result/Report.pm
```diff
@@ -377,15 +430,25 @@ sub fields_for_render {
...
 
-sub _read_security_marking {
-    my $self = shift;
+sub _read_security_marking
+{   my $self = shift;
 
     return $self->security_marking || $self->instance->read_security_marking;
 }
 
+sub group_ids
```
droberts-ctrlo, line 444: Only used within edit and add - otherwise, the search is within `Layout.pm` for group permissions for whom can view

lib/GADS.pm
```diff
@@ -2892,6 +2891,7 @@ prefix '/:layout_name' => sub {
                 my $checkbox_fields    = [body_parameters->get_all('checkboxes')];
                 my $security_marking   = body_parameters->get('security_marking');
                 my $instance           = $layout->instance_id;
+                my $groups             = body_parameters->get_all('groups');
```
abeverley, line 2894: This needs to be an array ref, otherwise every other group ID will become a key to this sub

lib/GADS.pm
```diff
@@ -2950,6 +2952,7 @@ prefix '/:layout_name' => sub {
                 my $checkboxes         = [body_parameters->get_all('checkboxes')];
                 my $security_marking   = body_parameters->get('security_marking');
                 my $instance           = $layout->instance_id;
+                my $groups             = body_parameters->get_all('groups');
```
abeverley, line 2955: Ditto for this

lib/GADS/Layout.pm
```diff
@@ -311,11 +311,30 @@ sub _build_security_marking {
 
 sub _build_reports
 {   my $self = shift;
-    my $reports_rs = $self->schema->resultset('Report')->search({
-        instance_id => $self->instance_id,
-        deleted => undef
-    });
```
abeverley, line 317: Keep the above, and then chain on the searches as required (or not in the case of superadmin)

lib/GADS/Layout.pm
```diff
@@ -311,11 +311,30 @@ sub _build_security_marking {
...
+            instance_id => $self->instance_id,
+            deleted => undef
+        });
+    } else {
+        $reports_rs = $self->schema->resultset('Report')->search({
+            instance_id => $self->instance_id,
+            deleted => undef,
+        },{prefetch => 'report_groups'})->search({
+            'report_groups.group_id' => { -in => \@group_ids },
+        });
```
abeverley, line 332: This is a bit messy. Better to just chain on the search from above with the relevant group IDs

abeverley:
This can be updated now for latest schema version

abeverley:
This all looks good, I think it just needs permission checking for the report download function (`GADS::Record::get_report()`, to ensure that users can only download the reports for the groups they are in. I would suggest adding a new function into `GADS::Schema::ResultSet::Report` in place of `find()` to ensure the user has access to the report and that it has not been deleted.

PR #441, opened 2024-07-22
Improve Testing Infrastructure

tomhukins:


droberts-ctrlo:
@tomhukins - a tiny bit of what was removed is in place for both cypress testing and local cypress testing - as I use Docker containers locally (that's personal preference, so I will admit that is my problem), which is why the host was set  (I tend to try to run tests locally when we have a "full" release) - I would also have assumed that the container used in the tests would also require this as it's a container for psql, although I'm open to being completely wrong, there :wink: . However, the basic user that was removed is there for testing within the Cypress tests (user permissions displaying correctly, etc).

@abeverley  I'll make another PR to add in the bits that were taken out that I believe need to be re-added - I shall also endeavour to ensure that things are commented on and noted where they may seem "unusual" to another party

tomhukins:
Thanks, @droberts-ctrlo, I hadn't thought about how these changes would affect anyone developing using containers.  I hope I haven't broken things too badly for you.  Ideally, we would have some kind of automated test to ensure such a development process works, but I'm sure it wouldn't be quick and easy to write one.

droberts-ctrlo:
@tomhukins—I will defer you to "not a chance" where testing is concerned—container infrastructure for other users is going to be left as an exercise for the user. It was bad enough getting a development container set up! :laughing: It's genuinely no problem at all. There's a new PR in, and I've made the "basic" user in the script reliant on an environment variable (as it is in the `seed_database.pl` script).

PR #448, opened 2024-08-07
Fixed form mapper error where File objects were not being mapped correctly

droberts-ctrlo:
We will now rely on `FormData` methods to catch erroneous objects being passed in. It may be worth working towards a more robust mapper for future use, but for now, this version is more than suitable for our needs.


src/frontend/js/lib/util/mapper/formdataMapper.ts
```diff
@@ -1,4 +1,5 @@
 export const formdataMapper = <T>(data: FormData | T) => {
+    console.log("DATA: ", data);
```
abeverley, line 2: Is this debug line still needed?

droberts-ctrlo: I just spotted that _literally_ as you put the comment in!

PR #449, opened 2024-08-07
File rename upload

droberts-ctrlo:
- Removed failing test
- Added definitions for filedrag plugin - this is to ease future development
- Added check for renaming in API endpoint as well as functionality to perform the rename
- Cleaned up some code within `DocumentComponent`, namely:
    - made `el` property `readonly` - this was because I believed it was being overwritten somewhere (see next point)
    - the `updateProgress` function wasn't firing correctly and couldn't find the element it was supposed to be linked to - this has been removed (although it now means the code isn't as DRY)
    - Changed the `showerror` function in an attempt to make result consistent across calls
- Added input and button for document renaming and implemented logic

As noted in the code - for some reason I wasn't able to modify the `original-name` data attribute within the component using `event.target` when the rename function was called so I had to pass the button in directly - frankly, TypeScript appears to have dropped the ball somewhere on this - I would upgrade the version of TS, Babel, Webpack, etc. but I'm not insane enough to go down that rabbit-hole - this is deemed the lesser of two evils in my opinion.

### DIFF
--- a/lib/GADS.pm
+++ b/lib/GADS.pm
@@ -1739,9 +1739,9 @@ post '/file/:id?' => require_login sub {
 # Use api route to ensure errors are returned as JSON
 post '/api/file/?' => require_login sub {
 
-    if(my rename_id=param('rename')) {
+    if(my $rename_id=param('rename')) {
         my $file = schema->resultset('Fileval')->find($rename_id);
-        $file->update({name=>param('name')});
+        $file->update({name=>param('filename')});
         return encode_json({
             id => $file->id,
             name => $file->name,
--- a/lib/GADS.pm
+++ b/lib/GADS.pm
@@ -1739,6 +1739,16 @@ post '/file/:id?' => require_login sub {
 # Use api route to ensure errors are returned as JSON
 post '/api/file/?' => require_login sub {
 
+    if(my rename_id=param('rename')) {
+        my $file = schema->resultset('Fileval')->find($rename_id);
+        $file->update({name=>param('name')});
+        return encode_json({
+            id => $file->id,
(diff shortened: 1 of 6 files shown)


lib/GADS.pm
```diff
@@ -1739,6 +1739,16 @@ post '/file/:id?' => require_login sub {
...
 post '/api/file/?' => require_login sub {
 
+    if(my $rename_id=param('rename')) {
+        my $file = schema->resultset('Fileval')->find($rename_id);
+        $file->update({name=>param('filename')});
+        return encode_json({
+            id => $file->id,
+            name => $file->name,
+            is_ok => 1,
+        });
```
abeverley, line 1749: The above enables anyone with a login to rename any file. You should be able to use `find_with_permission()` to control access.

Also, please update formatting :)

abeverley:
Thanks Dave, this looks good apart from the previous comment, although I am also thinking there needs to be a check to only allow renaming of files that haven't yet been attached to a record (to stop existing files being renamed). I'll also just need to test it in practice too.

lib/GADS.pm
droberts-ctrlo: Ironically, this code is defunct anyway - I moved it over to a put request and used that in #450 (so the request could be used in either case). I vote we close this PR and use #450 in its entirety(?)

abeverley: Yes thanks, sounds good, let's do all in #450. The same will apply - there will need to be checks to ensure that the renaming aspect can only be accessed accordingly.

droberts-ctrlo:
Closed as per comment in favour of #450

PR #450, opened 2024-08-08
Rename existing files

droberts-ctrlo:
Requires merging of #449 before merge

- Updated `datum::File` to check file name for marking datum as changed
- Removed rename within file post request and placed within a put request
- Updated `DocumentComponent` as required
- Removed debug statements
- Modified `edit.tt` as required


abeverley:
Thanks - this broadly looks fine, although there is a lot of duplicated code between the new `put` route and the existing `post` route. Can we combine some of this in a common function?

droberts-ctrlo:
> Thanks - this broadly looks fine, although there is a lot of duplicated code between the new `put` route and the existing `post` route. Can we combine some of this in a common function?

I'm not sure I can entirely see that—the only common code between the two seems to be where the file is created, and that's only (really) one statement. I will add that I may be completely wrong here, so I'll take another look anyway.

I'll add the security checks and the checks for a "new" record for now (although, in theory, the fact it's a new file should be passed as a parameter in the same way the delete option is in the `post` request).

droberts-ctrlo:
I don't know why the Firefox test is failing. This version contains the latest version of Dev as of 19/09/24 and the fixes from PR #459, which is why it has so many changes. Please merge #459 first, and I shall check the changes before merging to make sure nothing has fallen apart.

package.json
droberts-ctrlo, line 1: I needed to remove IE11, so removed unnecessary packages as well

src/frontend/components/form-group/input/lib/documentComponent.ts
droberts-ctrlo, line 1: There have been a _lot_ of changes here—some for efficiency, others because of a bug discovered that wasn't affected by the renaming changes, but it was felt the changes "fit" as they are within the same context.

src/frontend/components/form-group/input/lib/filenameChecker.ts
droberts-ctrlo, line 1: Although it is unknown whether this may be used in other components in the future, I was thinking of SRP more than anything.

src/frontend/js/lib/component.js
droberts-ctrlo, line 1: Input components have been refactored into separate components

src/frontend/js/lib/component.ts
droberts-ctrlo, line 1: This file has not changed since #459 - I cannot tell you why it is here!

Gotta love GitHub!

src/frontend/js/lib/component.test.ts
droberts-ctrlo, line 1: This file has not changed since #459 - I cannot tell you why it is here!

Gotta love GitHub!

src/frontend/js/lib/logging.js
droberts-ctrlo, line 1: Working towards a centralised logging solution - It is felt that this was a required change for testing

src/frontend/js/lib/util/domutils/lib/elementFactory.ts
droberts-ctrlo, line 1: I _believe_ this simplifies element creation, although I think it feels a little "dirty" - this is a utility function that will probably evolve over time, but is not felt to be a priority by any means

src/frontend/js/lib/util/filedrag/lib/filedrag.ts
droberts-ctrlo, line 1: Again, these changes are due to a bug on drag and drop that has only recently become apparent

views/edit.tt
droberts-ctrlo, line 1: The code in the JS and in the view differed in some ways - this has been modified for consistency

src/frontend/js/lib/util/filedrag/lib/filedrag.test.ts
droberts-ctrlo, line 1: Modified test to use proper way of calling events rather than using hacky casting.

src/frontend/components/form-group/input/lib/documentComponent.ts
droberts-ctrlo: Although I have realised that (obviously) a lot are also due to _the task being carried out_ - I clearly need coffee!

PR #452, opened 2024-08-08
Updated fullscreen to allow for filtering and searching

droberts-ctrlo:
Datatable was not being properly reinitialized within the code, an override parameter has been added to mitigate this.

### DIFF
--- a/src/frontend/components/data-table/lib/component.js
+++ b/src/frontend/components/data-table/lib/component.js
@@ -244,7 +244,6 @@ class DataTableComponent extends Component {
 
   // Self reference included due to scoping
   addSearchDropdown(column, id, index) {
-    console.log("Adding search dropdown");
     const $header = $(column.header())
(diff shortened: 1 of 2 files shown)


src/frontend/components/data-table/lib/component.js
```diff
@@ -244,6 +244,7 @@ class DataTableComponent extends Component {
 
   // Self reference included due to scoping
   addSearchDropdown(column, id, index) {
+    console.log("Adding search dropdown");
```
abeverley, line 247: Please remove debug line

abeverley:
Thanks Dave, looks good. The JS tests seem to be failing though - is this unrelated?

droberts-ctrlo:
That is due to some changes to the functionality I made a couple of pulls ago - although I couldn't tell you which PR it's in. I'll fix it here, and then we'll deal with the conflict later.

PR #457, opened 2024-09-13
Async Functions

droberts-ctrlo:
Updated functions where async would be more sensible to aid responsiveness


droberts-ctrlo:
I think I was a little premature, and overenthusiastic here with the removal of IE11 support - I want to close this for now and re-visit at a later time

PR #458, opened 2024-09-13
Common updates

droberts-ctrlo:
Refactored and updated `common.js` and usages

Requires merging of #457 first


droberts-ctrlo:
As previous (#457) has been closed, this one shall follow suit

PR #459, opened 2024-09-19
Updated component class to TS

droberts-ctrlo:
- Created basic `Component` test
- Updated and clarified Component class
- Modified and added type constraints to Component class
- Updated derived classes as necessary in order to ensure compilation

This branch requires merging _before_ any other PR currently awaiting merge


src/frontend/js/lib/component.ts
droberts-ctrlo, line 1: For a "proper" view of changes, diff this file with the original component.js

Typing and type checks are carried out by TSC most of the time (although there is some type erasure, but this is manageable), so some checks have been removed. Generics are also used to ensure that the correct type can be passed in—this is more for futureproofing (we may want to pass in an `HTMLButtonElement` at some point rather than a basic `HTMLElement`, for example). I believe most of this is ROC, or the comments/JSDoc provide relevant information.

PR #462, opened 2024-09-25
Fix for deleted reports remaining

droberts-ctrlo:
Deleted reports are still showing in system


abeverley:
Fixed with 5c232531 using `extant()`

PR #470, opened 2024-10-17
Fix for error in aggregate functions

droberts-ctrlo:
This branch is taken straight from v2.5.5 tag


droberts-ctrlo:
I didn't know which origin branch to do the PR for - obvs change it if you need to.

abeverley:
Thanks Dave. Just to note that you could also have done this if you wanted:
```
$data->{$_->id} = $agg->get_field_value($_)->for_table
    foreach grep {$_->aggregate}, @{$records->columns_render};
```

PR #476, opened 2024-11-19
Fixed menu-right code [B: 1623]

droberts-ctrlo:
- Modified classes to bypass Bootstrap code as this wasn't behaving as expected
- Added breakpoints to ensure alignment as expected


abeverley:
I'd prefer to investigate this further to see if we can fix natively using Bootstrap

droberts-ctrlo:
Would it be worth passing this across to the BS5 review and implementation (rather than doing twice)?

droberts-ctrlo:
This will be revisited on completion BS5 review as per comment above

PR #482, opened 2024-11-22
Fix for Typeaheads not working as expected with curvals

droberts-ctrlo:


src/frontend/components/data-table/lib/component.js
```diff
@@ -244,7 +244,7 @@ class DataTableComponent extends Component {
   }
 
   // Self reference included due to scoping
-  addSearchDropdown(column, id, index) {
+  async addSearchDropdown(column, id, index) {
```
droberts-ctrlo, line 247: As we're using fetch now

src/frontend/components/data-table/lib/component.js
```diff
@@ -290,21 +290,21 @@ class DataTableComponent extends Component {
...
     if (col.typeahead_use_id && searchValue) {
+      console.log("searchInput: ", $searchInput)
       $searchInput.after(`<input type="hidden" class="search">`)
-      $.ajax({
-        type: 'GET',
-        url: this.getApiEndpoint(columnId) + searchValue + '&use_id=1',
-        dataType: 'json'
-      }).done(function(data) {
-        if (!data.error) {
+      const response = await fetch(this.getApiEndpoint(columnId) + searchValue + '&use_id=1')
```
droberts-ctrlo, line 295: The AJAX request caused the table to refresh twice - I have opted to use Fetch as this allows us to await the response before rendering (`$.ajax` allows the function to return whilst still processing the request)

src/frontend/components/data-table/lib/component.js
```diff
@@ -321,7 +321,8 @@ class DataTableComponent extends Component {
             .withName(columnId.replace(/\s+/g, '') + 'Search')
             .withCallback((data) => {
               $('input', $header).val(data.name);
-              $('input', $header).trigger('change');
+              $('input.search', $header).val(data.id);
```
droberts-ctrlo, line 324: This fixes the problem on click of the value on the typeahead

droberts-ctrlo (commented):
General comments for the code as some changes may seem arbitrary but are required

src/frontend/components/data-table/lib/component.js
```diff
@@ -290,21 +290,21 @@ class DataTableComponent extends Component {
     const $searchInput = $(`<input class='form-control form-control-sm' type='text' placeholder='Search' value='${searchValue}'/>`)
     $searchInput.appendTo($('.input', $searchElement))
     if (col.typeahead_use_id && searchValue) {
+      console.log("searchInput: ", $searchInput)
```
abeverley, line 293: Looks good, I am just wondering if this console log can be removed?

droberts-ctrlo: Damn, thought I got them all! :wink:

PR #483, opened 2024-11-24
Implement auto-recover functionality [B: 1403]

droberts-ctrlo:


src/frontend/components/form-group/autosave/lib/autosaveBase.ts
```diff
@@ -0,0 +1,36 @@
...
+    get isClone() {
+        return location.search.includes('from');
+    }
+
+    get layoutId() {
+        return $('body').data('layout-identifier');
+    }
+
+    get recordId() {
+        const id = location.pathname.split('/').pop();
```
abeverley, line 19: I think it should be possible to get the record ID from the body of the edit page (e.g. `data-current-id` in the form)

lib/GADS.pm
```diff
@@ -4489,6 +4489,17 @@ prefix '/:layout_name' => sub {
 
 };
 
+get '/get_key' => require_login sub {
```
abeverley, line 4492: Consideration should be given to moving this to an `/api/` route in API.pm

lib/GADS.pm
```diff
@@ -4796,7 +4807,13 @@ sub _process_edit
         {
             # The "source" parameter is user input, make sure still valid
             my $source_curval = $layout->column(param('source'), permission => 'read');
-            try { $record->write(dry_run => 1, parent_curval => $source_curval) };
+            my %options = (dry_run => 1, parent_curval => $source_curval);
+            # If this is being submitted as part of an autosave recovery, allow
+            # mandatory values to be empty. This is because a field may have
+            # since been made mandatory, and we don't want this to cause the
+            # autosave recovery to fail
+            $options{force_mandatory} = 1 if param 'autosave';
```
abeverley, line 4815: My concern about this is that there is then a user way to always override mandatory (if they know what they are doing) - then can just add this parameter to any form submission. We probably need a way of tightening this, or preferably there would be a way of notifying the user that a saved value is now mandatory?

src/frontend/components/modal/modals/curval/index.js
```diff
@@ -1,10 +1,14 @@
...
-export default (scope) => {
-    if (!getComponentElements(scope, '.modal--curval').length) return;
+export default (scope) => initializeComponent(scope, '.modal--curval', CurvalModalComponent);
 
-    import(/* webpackChunkName="modal" */ "./lib/component")
-        .then(({ default: CurvalModalComponent }) => {
-            initializeComponent(scope, '.modal--curval', CurvalModalComponent)
-        });
-}
+//export default (scope) => {
```
abeverley, line 7: If this is no longer needed (?) can it be removed completely and/or comments added?

src/frontend/components/modal/modals/curval/lib/component.js
```diff
@@ -98,23 +171,14 @@ class CurvalModalComponent extends ModalComponent {
...
-            .html()
-        })
-        .join(", ")
-
-      guid = Guid()
+      guid ||= Guid()
       const id = `field${col_id}_${guid}`
       const deleteButton = multi
-        ? '<button class="close select-widget-value__delete" aria-hidden="true" aria-label="delete" title="delete" tabindex="-1">&times;</button>'
+        ? '<button class="close select-widget-value__delete" aria-hidden="true" aria-label="delete" title="delete" tabindex="-1">&times</button>'
```
abeverley, line 177: Should the semicolon have been removed?

abeverley (commented):
Great work Dave, this is looking really good. I've made a few minor comments so far. I'd also like to test it in practice, but I'm getting `Uncaught (in promise) TypeError: crypto.subtle is undefined` - any ideas?

src/frontend/components/form-group/autosave/lib/autosaveBase.ts
droberts-ctrlo: I can't find this - only the curval has this.

abeverley: Is it here or have I misunderstood? https://github.com/ctrlo/GADS/blob/3e003fb1460d24c0e57fba986b55ba4eb7af4336/views/edit.tt#L22

droberts-ctrlo: That's only on curvals.

I also didn't realise these comments are in realtime, I thought they were only when I did my push - that's useful to know! :smile:

abeverley: > That's only on curvals.

Fair enough, I hadn't spotted that, but maybe it could be moved outside of that condition? I haven't looked at it closely enough, but there should be something possible along those lines?

lib/GADS.pm
droberts-ctrlo: If I'm honest, I'm having some trouble with this one. I don't quite know whether it would be best to check the fields are mandatory (and blank) in the JS and provide feedback in the auto-recover modal or to have a better check (maybe using the `user_key` rather than just checking the param is 1) backend to ensure this is only called in the correct circumstances. Suggestions welcomed!!

droberts-ctrlo: Now handled in the JS

.github/workflows/tests.yml
```diff
@@ -178,7 +178,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [18.x, 20.x]
+        node-version: [18.x, 20.x, 22.x]
```
droberts-ctrlo, line 181: Update to use all currently supported node versions

.github/workflows/tests.yml
```diff
@@ -187,6 +187,7 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.node-version }}
-          cache: 'yarn'
+          cache: 'npm'
+      - run: npx update-browserslist-db@latest
```
droberts-ctrlo, line 191: Cache and browsers list for 18.x was "complaining"

jest.config.js
```diff
@@ -94,6 +94,9 @@ const config = {
     "^validation$": "<rootDir>/src/frontend/js/lib/validation",
     "^logging$": "<rootDir>/src/frontend/js/lib/logging",
     "^util/(.*)$": "<rootDir>/src/frontend/js/lib/util/$1",
+    "^components/(.*)$": "<rootDir>/src/frontend/components/$1",
```
droberts-ctrlo, line 97: Mapping was inconsistent between this file and `tsconfig.json`

lib/GADS/API.pm
```diff
@@ -1448,6 +1448,17 @@ any ['get', 'post'] => '/api/users' => require_any_role [qw/useradmin superadmin
     return encode_json $return;
 };
 
+get '/api/get_key' => require_login sub {
```
droberts-ctrlo, line 1451: Moved to API route for consistency

src/frontend/components/button/lib/cancel-button.ts
```diff
@@ -0,0 +1,15 @@
+import { clearSavedFormValues } from "./common";
+
+export default function createCancelButton(el: HTMLElement | JQuery<HTMLElement>) {
+    const $el = $(el);
+    if ($el[0].tagName !== 'BUTTON') return;
+    $el.data('cancel-button', "true");
+    $el.on('click', async () => {
+        const href = $el.data('href');
+        await clearSavedFormValues($el.closest('form'));        
+        if (href)
```
droberts-ctrlo, line 10: This should only work for buttons, and should have the HREF value set (to make the button act like a link where used) - if it isn't, just go back in the history.

src/frontend/components/button/lib/common.ts
droberts-ctrlo, line 1: This functionality is used in various buttons - it is abstracted out here to try to keep the code DRY

src/frontend/components/button/lib/rename-button.ts
```diff
@@ -187,12 +187,14 @@ class RenameButton {
...
 }
 
-(function ($) {
-    $.fn.renameButton = function () {
-        return this.each(function (_: unknown, el: HTMLButtonElement) {
-            new RenameButton(el);
-        });
-    };
-})(jQuery);
+if(typeof jQuery !== 'undefined') {
```
droberts-ctrlo, line 190: Some of the tests failed here. As this component isn't (currently) tested fully, it is thought that it is valid to remove the plugin functionality should jQuery be undefined.

src/frontend/components/form-group/autosave/index.js
```diff
@@ -0,0 +1,21 @@
+import { getComponentElements, initializeComponent } from 'component';
+import AutosaveComponent from './lib/component';
+import AutosaveModal from './lib/modal';
+import gadsStorage from 'util/gadsStorage';
+
+export default (scope) => {
+    try {
+        if (getComponentElements(scope, '.linkspace-field').length === 0) return;
```
droberts-ctrlo, line 8: Don't load the component if we don't have to

src/frontend/components/form-group/autosave/index.js
```diff
@@ -0,0 +1,21 @@
...
+        if (getComponentElements(scope, '#restoreValuesModal').length === 0) return;
+
+        if (gadsStorage.enabled) {
```
droberts-ctrlo, line 11: This is a shortcut to ensure `crypto.subtle` is available, if it isn't, we provide feedback, and remove the autorecover functionality.

src/frontend/components/form-group/input/lib/documentComponent.ts
```diff
@@ -1,4 +1,5 @@
 import 'components/button/lib/rename-button';
+import 'util/filedrag';
```
droberts-ctrlo, line 2: Code errors if this is left out

src/frontend/components/form-group/input/lib/documentComponent.ts
```diff
@@ -83,7 +84,7 @@ class DocumentComponent {
 
     async handleAjaxUpload(uri: string, csrf_token: string, file: File) {
         try {
-            if (!file) throw this.showException(new Error('No file provided'));
+            if (!file) this.showException(new Error('No file provided'));
```
droberts-ctrlo, line 87: Incorrect statement usage

src/frontend/components/form-group/input/lib/documentComponent.ts
```diff
@@ -181,5 +184,7 @@ class DocumentComponent {
  * @param {JQuery<HTMLElement> | HTMLElement} el The element to attach the document component to
  */
 export default function documentComponent(el: JQuery<HTMLElement> | HTMLElement) {
-    Promise.all([(new DocumentComponent(el)).init()]);
```
droberts-ctrlo, line 184: We actually need a component to be returned here now

src/frontend/js/lib/util/filedrag/index.ts
droberts-ctrlo, line 1: All that's changed here is the change to check for jQuery being defined so all the tests pass as expected

src/frontend/js/lib/util/gadsStorage/lib/AppStorage.ts
droberts-ctrlo, line 1: Abstract factory to create the relevant storage provider

src/frontend/js/lib/util/gadsStorage/lib/GadsStorage.test.ts
droberts-ctrlo, line 1: These tests are skipped on Node 18

src/frontend/components/form-group/autosave/index.js
droberts-ctrlo, line 1: Not entirely sure what the difference here is, but it works like this (as opposed to the previous version) so I'm not worried.

src/frontend/js/lib/util/common.test.ts
droberts-ctrlo, line 1: These tests are related to the common utility library and had to be changed due to the changes to said library.

**See the comment on the library for reasoning for this change being included here**

src/frontend/js/lib/util/common.ts
droberts-ctrlo, line 1: This involved importing the `globals.definitions` into the packaged code—this is **bad practice** and meant that some functions were being replaced by stubs and/or mocks.

Whilst this has not had any effect at this point, it *will* become a source of bugs and/or errors going forward, and required removal.

src/frontend/components/form-group/autosave/lib/component.js
```diff
@@ -0,0 +1,61 @@
...
+
+class AutosaveComponent extends AutosaveBase {
+  async initAutosave() {
+    const $field = $(this.element);
+    const self = this;
+    if ($field.data('is-readonly')) return;
+
+    // For each field, when it changes save the value to the local storage
+    $field.on('change', async function () {
+      const recovering = await self.storage.getItem('recovering');
```
droberts-ctrlo, line 12: Easiest way to ensure we're not writing data when we don't need to during a recovery

src/frontend/js/lib/util/gadsStorage/lib/GadsStorage.ts
```diff
@@ -0,0 +1,64 @@
...
+        const fetchResult = await fetch("/api/get_key");
+        const data = await fetchResult.json();
+        if (data.error !== 0) {
+            throw new Error("Failed to get storage key");
+        }
+        this.storageKey = data.key;
+    }
+
+    async setItem(key: string, value: string) {
+        if(await this.getItem('recovering')) return;
```
droberts-ctrlo, line 34: We turn off writing if we're performing a recovery to prevent extra write operations—this is more to prevent the odd curval error with dropdowns. 

It's felt it's more sensible to do this here, rather than search through all the code and try to work out where to put the check (and repeat it ad infinitum)

src/frontend/components/button/lib/common.ts
```diff
@@ -0,0 +1,32 @@
...
+        if (item) gadsStorage.removeItem(`linkspace-column-${field_id}-${layout}-${record}`);
+    }));
+}
+
+export function layoutId() {
+    return $('body').data('layout-identifier');
+}
+
+export function recordId() {
+    return isNaN(parseInt(location.pathname.split('/').pop())) ? 0 : parseInt(location.pathname.split('/').pop());
```
abeverley, line 23: Can we get this from the form via a data value?

lib/GADS/Schema/Result/User.pm
```diff
@@ -1226,4 +1228,23 @@ sub export_hash
...
 
+has encryption_key => (
+    is      => 'lazy',
+);
+
+sub _build_encryption_key {
+    my $self = shift;
+    
+    my $header_json  = '{"typ":"JWT","alg":"HS256"}';
+    # This is a string because encode_json created the JSON string in an arbitrary order and we need the same key _every time_!
```
abeverley, line 1239: There should be options in the JSON-encoder to order the keys, but happy to leave it like this for the time being.

lib/GADS/Schema/Result/User.pm
```diff
@@ -1226,4 +1228,23 @@ sub export_hash
...
+    my $payload_json = '{"sub":"' . $self->id . '","user":"' . $self->username . '"}';
+    my $header_b64   = encode_base64url($header_json);
+    my $payload_b64  = encode_base64url($payload_json);
+    my $input        = "$header_b64.$payload_b64";
+    my $secret       = sha256($self->password);
```
abeverley, line 1244: I assume it's been considered that a user changing their password will lose auto-recover data?

src/frontend/components/button/lib/common.ts
```diff
@@ -0,0 +1,32 @@
+import gadsStorage from "util/gadsStorage";
+
+export async function clearSavedFormValues($form: JQuery<HTMLElement>) {
```
abeverley, line 3: Just for my benefit, can you briefly comment what this function is used for and/or does?

src/frontend/components/button/lib/submit-draft-record-button.ts
```diff
@@ -1,13 +1,16 @@
...
  */
 export default function createSubmitDraftRecordButton(element: JQuery<HTMLElement>) {
-    element.on("click", (ev: JQuery.ClickEvent) => {
+    element.on("click", async (ev: JQuery.ClickEvent) => {
         const $button = $(ev.target).closest('button');
         const $form = $button.closest("form");
 
         // Remove the required attribute from hidden required dependent fields
         $form.find(".form-group *[aria-required]").removeAttr('required');
+        await clearSavedFormValues(ev.target.closest("form"));
```
abeverley, line 14: Ditto, please comment why saved form values are being cleared on save draft record

src/frontend/components/button/lib/submit-record-button.ts
```diff
@@ -46,6 +47,7 @@ export default class SubmitRecordButton {
                     if ($button.prop("name")) {
                         $button.after(`<input type="hidden" name="${$button.prop("name")}" value="${$button.val()}" />`);
                     }
+                    await clearSavedFormValues($form);
```
abeverley, line 50: Ditto :)

src/frontend/components/form-group/autosave/lib/autosaveBase.ts
```diff
@@ -0,0 +1,35 @@
...
+import gadsStorage from "util/gadsStorage";
+
+export default abstract class AutosaveBase extends Component {
+    constructor(element: HTMLElement) {
+        super(element);
+        this.initAutosave();
+    }
+
+    get isClone() {
+        return location.search.includes('from');
```
abeverley, line 11: I'd prefer to get this from some other method that doesn't rely on the URL, may data value again

src/frontend/js/lib/util/gadsStorage/lib/GadsStorage.ts
abeverley: Please add the above comment into the code as a comment

abeverley (commented):
Thanks Dave, almost there!!

lib/GADS/Schema/Result/User.pm
droberts-ctrlo: It has - this is basically because we were trying to work out a finite time before data becomes "stale" and we decided this was the best approach.

PR #484, opened 2024-11-25
Addition of deleted date to person field [B: 1742]

droberts-ctrlo:


abeverley:
I think we discussed this - it should be possible to make some improvements by returning the DateTime object and then obtaining the epoch from that, rather than creating a new function.

droberts-ctrlo:
> I think we discussed this - it should be possible to make some improvements by returning the DateTime object and then obtaining the epoch from that, rather than creating a new function.

We did - it's on my list. :smile:

PR #485, opened 2024-11-25
Updated ts to ignore errors

droberts-ctrlo:


droberts-ctrlo:
Basically, some of the Digitpaint stuff that was in for their testing wasn't properly "type-safe" - I've told the compiler to ignore it for now. Why it would only just become an issue now, God only knows!

PR #487, opened 2024-11-25
Updated curvals to work as expected

droberts-ctrlo:


src/frontend/components/data-table/lib/component.js
```diff
@@ -289,18 +290,19 @@ class DataTableComponent extends Component {
      * insertion into the visible input */
     const $searchInput = $(`<input class='form-control form-control-sm' type='text' placeholder='Search' value='${searchValue}'/>`)
     $searchInput.appendTo($('.input', $searchElement))
-    if (col.typeahead_use_id && searchValue) {
+    if (col.typeahead_use_id) {
       $searchInput.after(`<input type="hidden" class="search">`)
+      if(searchValue) {
```
abeverley, line 295: The indentation looks wrong here, but I'll check once I've merged and fix if required

PR #488, opened 2024-11-26
Remove ordering from metadata

droberts-ctrlo:


droberts-ctrlo:
After review process, this has been decided that this will not solve the underlying problem, and may also cause issues with ordering by other fields outside of curvals. Closing in favour of creating a more comprehensive and correct solution.

PR #489, opened 2024-11-27
Made "view all" link optional in all curvals [B: 1734]

droberts-ctrlo:


droberts-ctrlo:
With a view to comments on B1784, it has been decided that this request be closed in favour of a different solution

PR #490, opened 2024-11-29
Fix for as_integer causing panic [B: 1741]

droberts-ctrlo:
Due to extensions using hashinflater to create reports, the panic on `GADS::Datum::Integer->as_integer` was triggering on generating reports for specific systems - this has been put back in with a view to fix this issue.


abeverley:
Closing pending different fix

PR #493, opened 2024-12-02
Reject reason complete [B: 1647]

droberts-ctrlo:
Initial commit contained "odd" changes due to merge of current dev branch - this is taken from the current dev branch (post v2.5.11).

### DIFF
--- a/lib/GADS.pm
+++ b/lib/GADS.pm
@@ -1556,8 +1556,9 @@ any ['get', 'post'] => '/user_requests/' => require_any_role [qw/useradmin super
                 if logged_in_user->id == $delete_id;
 
         my $usero = rset('User')->find($delete_id);
+        my $email_reject_text = param('reject_reason');
 
-        if (process( sub { $usero->retire(send_reject_email => 1) }))
+        if (process( sub { $usero->retire(send_reject_email => 1, email_reject_text => $email_reject_text) }))
         {
             $audit->login_change("User ID $delete_id deleted");
             return forwardHome(
--- a/lib/GADS/Schema/Result/User.pm
+++ b/lib/GADS/Schema/Result/User.pm
@@ -1015,22 +1015,19 @@ sub permissions
 }
 
 sub _map_fields {
-  my ($self, $text) = @_;
-  my @fields = ('firstname', 'surname', 'email', 'title', 'organisation', 'department', 'team');
+    my ($self, $text) = @_;
+    my @fields = ('firstname', 'surname', 'email', 'title', 'organisation', 'department', 'team');
   
-  if($text) {
-    foreach my $field (@fields) {
-      my $value = $self->$field || '';
-      $text =~ s/\{$field\}/$value/g;
+    if($text) {
+        foreach my $field (@fields) {
+            my $value = $self->$field || '';
+            $text =~ s/\{$field\}/$value/g;
+        }
+        my $notes = $self->account_request_notes || '';
+        $text =~ s/\{notes\}/$notes/g;
     }
-  }
 
-  if($text) {
-    my $notes = $self->account_request_notes || '';
-    $text =~ s/\{notes\}/$notes/g;
-  }
-
-  return $text;
+    return $text;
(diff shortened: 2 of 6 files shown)


lib/GADS.pm
```diff
@@ -1556,8 +1556,9 @@ any ['get', 'post'] => '/user_requests/' => require_any_role [qw/useradmin super
                 if logged_in_user->id == $delete_id;
 
         my $usero = rset('User')->find($delete_id);
+        my $email_reject_text = param('reject_reason');
 
-        if (process( sub { $usero->retire(send_reject_email => 1) }))
+        if (process( sub { $usero->retire(send_reject_email => 1, email_reject_text => $email_reject_text) }))
```
droberts-ctrlo, line 1561: I assume that if the reject text is blank, the function being called will treat it as nil (or false)?

lib/GADS/Schema/Result/User.pm
```diff
@@ -1014,6 +1014,25 @@ sub permissions
     }
 }
 
+sub _map_fields {
```
droberts-ctrlo, line 1017: A bit of "sugar" - saves users having to type in fields and/or remember values.

src/frontend/components/data-table/lib/component.js
```diff
@@ -8,7 +8,6 @@ import 'datatables.net-rowreorder-bs4'
 import { setupDisclosureWidgets, onDisclosureClick } from 'components/more-less/lib/disclosure-widgets'
 import { moreLess } from 'components/more-less/lib/more-less'
 import { bindToggleTableClickHandlers } from './toggle-table'
-import TypeaheadBuilder from 'util/typeahead'
```
droberts-ctrlo, line 11: This isn't related (strictly speaking) but needs removing, and as I was working on the file :smiley:

src/frontend/components/data-table/lib/component.js
```diff
@@ -126,7 +125,7 @@ class DataTableComponent extends Component {
 
   fillModalData(row) {
     const fields = $(this.modal).find('input, textarea')
-    const btnReject = $(this.modal).find('.btn-js-reject-request')
+    const btnReject = $(this.modal).find('.btn-js-reject-request-send')
```
droberts-ctrlo, line 128: Had to give this a different name as not to end up with clashing data

src/frontend/components/modal/lib/component.js
```diff
@@ -271,6 +271,7 @@ class ModalComponent extends Component {
 
   // Validate the required fields of the frame
   validateFrame() {
+    if(!this.frame) return;
```
droberts-ctrlo, line 274: This is an "odd" bug that's just arrived - I don't know if it was present before, but if the frame hasn't been visible, it is set to undefined/null, therefore setting invalid was throwing an exception.

views/wizard/user_add.tt
```diff
@@ -29,9 +29,9 @@
 
           <ol class="modal__steps">
             <li class="modal__step modal__step--active" data-step="1">
-                  <span>
```
droberts-ctrlo, line 32: Just some odd indentation I spotted and fixed.

views/wizard/user_add.tt
```diff
@@ -292,6 +292,53 @@
               }];
           -%]
         </div>
+        <div class="modal-frame" data-config='{"step":4,"frame":4,"item":null,"count":null}'>
```
droberts-ctrlo, line 295: Whole new frame for delete reason - this isn't in the "list" on the header to avoid confusion when creating an account.

lib/GADS.pm
abeverley: Yes, correct.

abeverley:
Thanks Dave, this broadly looks fine, pending UX testing. There are a few formatting issues (e.g. tabs should be 4 spaces, curly brackets on new line)

droberts-ctrlo:
UX Review complete - ready for code review

PR #494, opened 2024-12-02
Fix for user notes not persisting when saving users [B: 1736]

droberts-ctrlo:


abeverley:
This looks like there could be some improvements to data structure, to avoid different variable names doing the same thing. But that can follow another time.

droberts-ctrlo:
> This looks like there could be some improvements to data structure, to avoid different variable names doing the same thing. But that can follow another time.

I agree, if I'm honest (in retrospect and looking at the code) - I may look at this a bit more when we're in the change-freeze if nothing else needs doing. :smile:

PR #497, opened 2024-12-03
Removed webdriver chromium test due to failure [B NONE]

droberts-ctrlo:
No B for this one - Erroring test is failing for an unknown reason - removing Chromium tests for now as (most) Cypress tests should cover what's missed with Chromium.


droberts-ctrlo:
Shortcut - fixing the test is a better solution; I'll stop being lazy :wink:

PR #502, opened 2024-12-18
Added nil checks on person deleted field

droberts-ctrlo:


abeverley (commented):
Thanks Dave. I think better to do a check on `$_->{deleted}` being null, otherwise you're calling `parse_datetime` twice.

PR #503, opened 2024-12-19
Timeline triage

droberts-ctrlo:
This PR is purely to address the timeline making the system unresponsive. The fix:

- Sets an (arbitrary) limit on the number of items one request can return - in testing on a copy system it was found that this gave us an average wait time of approx 1min when reaching said limit
- Runs the request to update the timeline asynchronously, meaning the system remains responsive, and navigation away from the page is still possible should the timeline be selected by mistake
- Cancels requests should another be required before it has completed

As stated, this is currently a triage operation, this is not expected to be a permanent fix. A full requirement shall be discussed and implemented for v2.7 as far as is feasible.

### DIFF
--- a/lib/GADS/Timeline.pm
+++ b/lib/GADS/Timeline.pm
@@ -428,6 +428,11 @@ sub _build_items
 
         $self->_set_display_to($oldest)
             if $oldest > ($self->display_to || AT_BIGBANG);
+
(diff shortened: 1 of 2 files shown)


droberts-ctrlo:
Closing to persue a better solution

PR #507, opened 2025-01-21
Refactored code and added await to stop race conditions

droberts-ctrlo:


src/frontend/components/form-group/autosave/lib/modal.js
```diff
@@ -11,24 +11,25 @@ class AutosaveModal extends AutosaveBase {
...
-    
+
     $modal.find('.btn-js-restore-values').on('click', async (e) => {
-      this.storage.setItem('recovering', true);
-      let curvalCount = $form.find('.linkspace-field[data-column-type="curval"]').length;
-      
       e.preventDefault();
+      e.stopPropagation();
+
+      await this.storage.setItem('recovering', true);
```
droberts-ctrlo, line 19: This was causing a race condition that was borken

src/frontend/components/form-group/autosave/lib/modal.js
```diff
@@ -11,24 +11,25 @@ class AutosaveModal extends AutosaveBase {
...
+      let curvalCount = $form.find('.linkspace-field[data-column-type="curval"]').length;
 
       let errored = false;
 
       let $list = $("<ul></ul>");
       const $body = $modal.find(".modal-body");
       $body.html("<p>Restoring values...</p>").append($list);
-      await Promise.all($form.find('.linkspace-field').map(async (_,field) => {
```
droberts-ctrlo, line 26: Don't need this to be awaited because the finally (after this) encapsulates what would happen at the end either way.

src/frontend/components/form-group/autosave/lib/modal.js
```diff
@@ -11,24 +11,25 @@ class AutosaveModal extends AutosaveBase {
...
+      Promise.all($form.find('.linkspace-field').map(async (_, field) => {
         const $field = $(field);
-        await this.storage.getItem(this.columnKey($field)).then(json => {
+        try {
+          const json = await this.storage.getItem(this.columnKey($field))
```
droberts-ctrlo, line 30: I was being a little overzealous with my usage of `Promise` objects here - IMO this is "cleaner" now; we still need the `Promise.all` because the field mappings need to run decryption asynchronously.

droberts-ctrlo (commented):
This appears to behave "better" - looks like the setting of the `recovering` key wasn't always returning as expected (as it wasn't awaited) but I feel it may need checking further in a test environment just to be certain (local testing works fine).

src/frontend/components/form-group/autosave/lib/modal.js
```diff
@@ -41,43 +42,44 @@ class AutosaveModal extends AutosaveBase {
...
+                if (!curvalCount) this.storage.removeItem('recovering');
               });
             }
             setFieldValues($field, values);
-            if(type !== "curval") {
+            if (type !== "curval") {
               const $li = $(`<li class="li-success">Restored ${name}</li>`);
               $list.append($li);
             }
+            // For some reason a delay is needed here else the promise forgets what it's doing frontend
```
abeverley, line 59: Is this comment still needed?

abeverley (commented):
Thanks Dave, looks good, can you just add some comments inline to the code?

PR #508, opened 2025-01-23
Autosave fix

droberts-ctrlo:
Fixes for Autosave not completely working correctly on curvals

### DIFF
--- a/src/frontend/components/modal/modals/curval/lib/component.js
+++ b/src/frontend/components/modal/modals/curval/lib/component.js
@@ -5,6 +5,7 @@ import { guid as Guid } from "guid"
 import { initializeRegisteredComponents } from 'component'
 import { validateRadioGroup, validateCheckboxGroup } from 'validation'
 import gadsStorage from 'util/gadsStorage'
+import { fromJson } from 'util/common'
 
 class CurvalModalComponent extends ModalComponent {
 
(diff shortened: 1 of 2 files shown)


src/frontend/components/modal/modals/curval/lib/component.js
```diff
@@ -333,6 +335,9 @@ class CurvalModalComponent extends ModalComponent {
     let url = current_id
       ? `/record/${current_id}`
       : `/${instance_name}/record/`
+    
+    if(url.endsWith('undefined'))
+      url=`/${instance_name}/record/`
```
abeverley, line 340: Must be a better way of doing this surely..?

abeverley (commented):
Thanks Dave. Could you just remove the config statements, plus the separate comment please?

src/frontend/components/modal/modals/curval/lib/component.js
droberts-ctrlo: I found if I put

```javascript
let url = current_id || current_id !== 'undefined'
    ? `record/${current_id}
    : `/${instance_name}/record`;
```

It doesn't allow you to recover the values in the curval for some reason (the logic is correct afaik) but with the 2-stage check, it works fine.

abeverley: I'm more thinking as to where the source of the string 'undefined' is coming from. At the point of ```let url = current_id...``` is current_id a string consisting of the text 'undefined'? If so, would it be possible to find the source of that and make sure that it's actually an undefined variable being passed in rather than the string 'undefined'?

PR #512, opened 2025-01-29
Made timeline async

droberts-ctrlo:
Rendering of timeline is now async in order to ensure requests don't lock up the system and navigation can still be performed away from the timeline


abeverley:
Thanks Dave. There are a lot of changed lines, but I assume that is just because of the indent. Can you just confirm it is essentially only adding the promise?

PR #513, opened 2025-01-31
Added fix for timeout on autorecover [B:1361]

droberts-ctrlo:
- Added a record identifier to the session
- Removed clear from the save button code (we want the data cleared on successful submission)
- Created a new Alert component that checks for the data attribute that contains the record ID, and clears the data using that if it's present
- Added data attribute to the message_center view


abeverley:
Despite me saying that this should go in the alert, I think that actually on reflection it can be kept separate (I was originally thinking that using the alert was the only way of keeping it persistent across requests). I think that actually you should set a session variable on successful submit (as you are doing), but instead it should be a more generic variable (e.g. `actions_for_page`). This would then be rendered as data in the `body` of the page (or similar) and passed using the before_template hook. Then, the Javascript should be updated to look for those particular data items and act accordingly. This could potentially be used for other actions we need the Javascript to do on page load.

As such, I think it should be rendered for all pages, not just the data page.

Finally, having seen the complicated code needed for removing all the required fields, I am wondering if we can change the data structure of the local storage, having a way of grouping all fields together for one instance, so that they can be easily deleted rather than needing to loop through and find each one.

views/snippets/message_center.tt
```diff
@@ -1,6 +1,6 @@
 [% IF messages.size %]
   [% FOR message IN messages %]
-  <div class="alert alert-[% message.bootstrap_color %]" role="alert">
+  <div class="alert alert-[% message.bootstrap_color %]" role="alert"[% IF record_id.defined %] data-record-id=[% record_id %][% END %]>
```
abeverley, line 3: Not sure if this change is still needed?

views/layouts/main.tt
```diff
@@ -9,7 +9,7 @@
     <link rel="stylesheet" href="[% url.css %]/external.css?v=7">
     <link rel="stylesheet" href="[% url.css %]/general.css?v=8">
   </head>
-  <body class="[% body_class || 'page' %]"[% IF csrf_token %] data-csrf="[%- csrf_token -%]"[% END; IF layout OR layout_obj; %] data-layout-identifier="[% layout.identifier || layout_obj.identifier %]"[% END %]>
+  <body class="[% body_class || 'page' %]"[% IF csrf_token %] data-csrf="[%- csrf_token -%]"[% END; IF layout OR layout_obj; %] data-layout-identifier="[% layout.identifier || layout_obj.identifier %]"[% END %][% IF actions; %][% IF actions.update_record.defined %] data-update-record="[% actions.update_record %]"[% END %][% END; %]>
```
abeverley, line 12: I'm wondering whether the actions object can be passed into the body data as its full object rather than splitting out a particular action?

lib/GADS.pm
```diff
@@ -4801,6 +4805,9 @@ sub _process_edit
                 my $forward = (!$id && $layout->forward_record_after_create) || param('submit') eq 'submit-and-remain'
                     ? 'record/'.$record->current_id
                     : $layout->identifier.'/data';
+                $actions->{update_record} = $id ? $id: 0;
```
abeverley, line 4808: I think this should maybe be called clear_saved_values or similar instead of update_record?

abeverley (commented):
Thanks Dave, just some minor comments for consideration

PR #515, opened 2025-02-06
Fix for similar users on multiple tenants [B:1625]

droberts-ctrlo:


abeverley:
Thanks Dave. I can see that this would work as fix, but I think the wider problem is that site_id is not set at this point (which should then automatically be used) https://github.com/ctrlo/GADS/blob/3e003fb1460d24c0e57fba986b55ba4eb7af4336/lib/GADS/Schema.pm#L19

Can you see if you can work out why site_id is not being set?

abeverley:
Thanks Dave, looks good. Please just remove the Data::Dump module and move the `and` back to the 2nd line (where I feel it's clearer as to what is happening)

PR #516, opened 2025-02-14
Made curval modal get values when editing from fields

droberts-ctrlo:
The code will load the curval values if they have to from the backend, otherwise, will pull values from the field as defined in the hidden value


droberts-ctrlo:
This branch is going to be closed and re-done with a post request if possible

PR #520, opened 2025-03-05
Added code to truncate select fields in filters when selected to max 50 [B:1785]

droberts-ctrlo:


droberts-ctrlo:
Due to this being a user issue, and truncating text in this manner will effect screen-readers, this PR is closed.

PR #521, opened 2025-03-05
Updated curval filter to not use custom typeahead for value [B: 1787]

droberts-ctrlo:


abeverley:
I'm not sure this is the right solution as it's removing functionality (unreferenced in related ticket). It would be better to fix the functionality, allowing typeahead or custom values as required.

PR #522, opened 2025-03-07
Updated code to ensure indexing on values when they are in return value fields is correct [B: 1765]

droberts-ctrlo:


abeverley:
As discussed please consider combining the 2 blocks of code that deal with a similar problem

PR #523, opened 2025-03-10
Fix for chronology on curvals [B:1783]

droberts-ctrlo:
This fix means the curvals will stop showing as added or removed when they "drop off" the view due to view limits.


droberts-ctrlo:
Closed in favour of different fix as implemented by @abeverley

PR #527, opened 2025-03-21
Ensured error only shows once when encryption is disabled

droberts-ctrlo:


abeverley:
I'm happy to merge this, although in general I think I'd prefer a specific class on the warning and then just searching to see if that exists before adding.

PR #531, opened 2025-03-31
Updated jest config

droberts-ctrlo:
- Updated configuration
- Added further tests for buttons and React components

PR #537, opened 2025-04-15
Standardizes code formatting and import style

droberts-ctrlo:
Purely updating code style and formatting - relies on #531 being merged first


droberts-ctrlo:
As with previous request - this is becoming messy - will redo with single PR as far as possible

PR #539, opened 2025-04-16
Improves and fixes various components

droberts-ctrlo:
Relies on #538


droberts-ctrlo:
Closing in favour of cleaner solution

PR #541, opened 2025-04-16
Upgrades to basic JS library packages

droberts-ctrlo:
Relies on #539


droberts-ctrlo:
Closing with a view to having a cleaner commit tree

PR #542, opened 2025-04-16
Upgrade React to v19

droberts-ctrlo:
Relies on #541


droberts-ctrlo:
I have erroneously included the BS5 changes within this set of commits - I shall close this one and rollback said commits and then re-open the PR with the correct changes

PR #543, opened 2025-04-17
React upgrades

droberts-ctrlo:
Relies on #541


droberts-ctrlo:
Closing in favour of cleaner commit tree

PR #544, opened 2025-04-24
Rolled back error causing filter typeaheads not to work as expected [B: 1815]

droberts-ctrlo:
Filter typeaheads were returning an error when trying to save views using them on curvals - this error was caused by #521

In testing, the above caused issues when using short names for filtering on curvals. This issue has been resolved another way.


abeverley:
Thanks - I'm wondering if it's possible to add a Cyprus test too?

PR #546, opened 2025-05-01
Updated component and controller code [B:1767]

droberts-ctrlo:
- Updated controller to show an error on session expiry rather than generic message
- Updated component to display error message rather than generic one if there is a message to display, otherwise a generic message is given


droberts-ctrlo:
Requires #533 merging first

PR #548, opened 2025-05-20
Updated code to make GET requests where data is exchanged to use POST [B:1533, 1763]

droberts-ctrlo:


lib/GADS/API.pm
```diff
@@ -420,7 +420,7 @@ post '/api/token' => sub {
 
 prefix '/:layout_name' => sub {
 
-    get '/api/field/values/:id' => require_login sub {
+    any ['get', 'post'] => '/api/field/values/:id' => require_login sub {
```
abeverley, line 423: I'm wondering whether this can now just be a POST route?

src/frontend/components/modal/modals/curval/lib/component.js
```diff
@@ -330,14 +333,13 @@ class CurvalModalComponent extends ModalComponent {
...
-  getURL(current_id, instance_name, layout_id, form_data) {
+  getURL(current_id, instance_name, layout_id) {
 
     let url = current_id
       ? `/record/${current_id}`
       : `/${instance_name}/record/`
 
-    url = `${url}?include_draft&modal=${layout_id}`
-    if (form_data) url = url + `&${form_data}`
+    url += `?csrf-token=${$('body').data("csrf")}&modal=${layout_id}&include_draft=1`
```
abeverley, line 342: I don't think the CSRF token needs to be in the query - it should suffice just in the body (as presumably it already is)

src/frontend/js/lib/util/common.test.ts
```diff
@@ -1,5 +1,5 @@
 import "../../../testing/globals.definitions";
-import { fromJson, hideElement, showElement } from "./common";
+import { fromJson, hideElement, showElement, urlDataToJson } from "./common";
```
abeverley, line 2: I don't think this change is needed anymore

src/frontend/js/lib/util/common.ts
```diff
@@ -1,13 +1,18 @@
+export const isDefined = <T>(value: unknown): value is T => value !== undefined && value !== null;
+export const isString = (value: unknown): value is string => isDefined(value) && typeof value === 'string';
+export const isNotEmptyString = (value: unknown): value is string => isString(value) && value.trim() !== '';
+export const isEmptyString = (value: unknown): value is string => isString(value) && value.trim() === '';
```
abeverley, line 4: I think changes to this file can be removed completely?

src/frontend/js/lib/util/mapper/formdataMapper.ts
```diff
@@ -11,6 +11,7 @@ export const formdataMapper = <T>(data: FormData | T) => {
     if(data instanceof Object && Object.keys(data).length === 0) throw new Error("Cannot map an empty object");
     const formData = new FormData();
     Object.entries(data).forEach(([key, value]) => {
+        console.log("Adding to formData", key, value);
```
abeverley, line 14: No longer needed?

abeverley (commented):
Thanks Dave, just a few minor comments

lib/GADS/API.pm
droberts-ctrlo: I left this as a GET/POST because I didn't know if there would likely be anywhere it's still used as a GET that I may have missed, and thus this would break the application. I am all for making it exclusively POST, but I do wonder if GET should be kept in to make sure there are no edge-cases missed.

src/frontend/js/lib/util/common.ts
droberts-ctrlo: I forgot they were there, TBH - useful utility methods, but in this instance, they are neither required, nor related to this PR.

PR #549, opened 2025-05-23
Modified auth check for token requests to properly use Basic auth [B:1832]

droberts-ctrlo:


droberts-ctrlo:
Fixed with commit ca529ea0

PR #550, opened 2025-06-10
Enhances file upload flexibility

droberts-ctrlo:
Adds functionality to allow administrators to configure allowed file extensions and MIME types via a script.


lib/GADS/Role/Presentation/Column/File.pm
```diff
@@ -0,0 +1,11 @@
+package GADS::Role::Presentation::Column::File;
+
+use Moo::Role;
+
+sub after_presentation
+{   my ($self, $return) = @_;
+
+    $return->{extra_values} = $self->extra_values;
```
abeverley, line 8: I think the part of this PR which involves rendering the allowed mime types in the front-end and then returning it would be better to all be done in the back end.

lib/GADS.pm
```diff
@@ -1796,7 +1796,17 @@ post '/api/file/?' => require_login sub {
 
     if (my $upload = upload('file'))
     {
-        my $mimetype = $filecheck->check_file($upload, check_name => 0); # Borks on invalid file type
+        my $user = logged_in_user;
+        my $column_id = body_parameters->get('column_id') // 0;
```
abeverley, line 1800: As there should always be a column ID (?) better to do:
```
my $column_id = body_parameters->get('column_id')
    or error __"Missing column ID";
```

lib/GADS/Filecheck.pm
```diff
@@ -29,46 +29,67 @@ sub check_file
...
 
     # As recommended at https://owasp.org/www-community/vulnerabilities/Unrestricted_File_Upload
     # Brackets have been added to this - above recommendations do not explicitly state that brackets are not allowed - Ticket #1695
     $self->check_name($upload->filename)
         if($check_name);
 
-    error __"Maximum file size is 5 MB"
-        if $upload->size > 5 * 1024 * 1024;
+    error __"Maximum file size is 50 MB"
+        if $upload->size > 50 * 1024 * 1024;
```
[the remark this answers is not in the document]
droberts-ctrlo: The original file size check was here - I've kept it in, rather than allow (theoretically) infinite sizes. We settled on a maximum of 50MB as a "sensible" limit.

lib/GADS/Filecheck.pm
```diff
@@ -29,46 +29,67 @@ sub check_file
 {   my ($self, $upload, %options) = @_;
 
     my $check_name = $options{check_name} // 1;
+    my $allowed_data = $options{extra_types};
 
+    my $mimeRegex = '^(' . join('|', map { $_->{name} } @$allowed_data) . ')' if $allowed_data && @$allowed_data;
+    my $filetypeRegex = '^(' . join('|', map {$_->{extension}} @$allowed_data) . ')$' if $allowed_data && @$allowed_data;
```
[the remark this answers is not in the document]
abeverley: Not sure that this has been resolved?

lib/GADS/Column/File.pm
```diff
@@ -21,10 +21,31 @@ package GADS::Column::File;
...
 
 extends 'GADS::Column';
 
+has '+option_names' => (
+    default => sub { [qw/override_types/] }
+);
+
+has override_types => (
+    is      => 'rw',
+    isa     => Maybe[ArrayRef],
```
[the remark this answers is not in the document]
abeverley: I've seen the comment about it being undef for no options, but surely with the update to the builder function it will be an empty arrayref if no options?

bin/add-extension.pl
```diff
@@ -0,0 +1,98 @@
...
+GetOptions(
+    'mimetype=s'  => \$mimetype,
+    'extension=s' => \$extension,
+    'instance=i'  => \$instance,
+    'layout=i'    => \$layout_id
+);
+
+unless ( $mimetype && $extension && $instance && $layout_id ) {
+    say "Usage $0 --mimetype=<mimetype> --extension=<extension> --layout=<layout_id> --instance=<instance_id>";
+    exit 1;
```
abeverley, line 36: Just do `error "Usage..."` and then there is no need for the exit

bin/add-extension.pl
```diff
@@ -0,0 +1,98 @@
...
+}
+
+# we need a config instance, and a filecheck instance
+GADS::Config->instance( config => config );
+my $file_check = GADS::Filecheck->instance();
+
+# Check against the already allowed mime-types and error if it already exists
+if ( $file_check->check_type($mimetype) ) {
+    error __x"Mimetype {type} already allowed", type=>$mimetype;
+    exit 2;
```
abeverley, line 46: No need for the exit - the call to error will exit

bin/add-extension.pl
```diff
@@ -0,0 +1,98 @@
...
+    schema                   => schema,
+);
+
+# Get the column
+my $result     = $layout->column($layout_id);
+
+# Check if the column is of the correct type - there's no point setting this property on the wrong one!
+if ( $result->type ne 'file' ) {
+    error __x"Invalid column for properties to be set on";
+    exit 4;
```
abeverley, line 74: Ditto

lib/GADS/Filecheck.pm
```diff
@@ -25,50 +25,109 @@ sub is_image
...
+{   my ($self, $upload, %options) = @_;
+
+    my $check_name = $options{check_name} // 1;
+    my $allowed_data = $options{extra_types};
+
+    my ($mimeRegex, $filetypeRegex) = $self->create_regex($allowed_data);
+    
+    my $info = $magic->info_from_filename($upload->tempname);
+
+    my $allowed_bypass = $info->{mime_type} =~ qr/$mimeRegex/ if $mimeRegex;
```
abeverley, line 82: May as well return the regex from `create_regex` as the function name implies

PR #553, opened 2025-07-02
Amended to allow for request that was missed previously [B:1533]

droberts-ctrlo:


droberts-ctrlo:
Required to keep the "custom" filtering functionality due to Typeaheads causing UX issues and display becomes confusing

PR #554, opened 2025-07-16
Updated config to allow for migration to take `DANCER_CONFDIR` variable into account

droberts-ctrlo:


droberts-ctrlo:
Incorrect commit - will close

PR #556, opened 2025-07-21
Fixes for file uploads

droberts-ctrlo:
Fixes include:
- User option fixes to annotate options that Users can't modify
- Checks to file types that check overrides on submission to prevent spoofing
- Changes to filecheck to check files on filesystem on submission rather than just uploads
- Minor fix to upload progress display

### DIFF
--- a/bin/add-extension.pl
+++ b/bin/add-extension.pl
@@ -72,6 +72,7 @@ if ( $result->type ne 'file' ) {
 # Make sure the user wants to add the data to the selected column.
 print "Do you want to add mime type $mimetype with extension $extension to ". $result->name ." (y/n): ";
 my $answer = <STDIN>;
+chomp $answer;
 
 # Do a check for y
 if ( $answer =~ /^y/i ) {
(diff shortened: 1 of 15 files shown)


abeverley:
I had to add this line in cee83a44e6 to fix failing tests (this PR has caused the unit tests to fail again). I'm wondering what the problem is that this is fixing?

droberts-ctrlo:
> I had to add this line in [cee83a4](https://github.com/ctrlo/GADS/commit/cee83a44e606ea51348b341d02c8dd0571e2ed57) to fix failing tests (this PR has caused the unit tests to fail again). I'm wondering what the problem is that this is fixing?

I'll be honest - I need to go through some of the new upload "stuff" anyway - it's currently not possible to create a document field, or upload files (in the "files" menu on the left). This fixes the latter, but not the former - I will go back to this and use this PR to implement all fixes for any upload issues, as well as fix unit tests and let you know when all issues are resolved (and, therefore, this can be merged)

PR #559, opened 2025-07-29
Updated changelist and version file for v2.7.0

droberts-ctrlo:
Relies on #556, #557, and #558


droberts-ctrlo:
Will redo ensuring only changes to changelog are present

PR #563, opened 2025-08-01
Fixes to allow tests to run within containers [B:1852]

droberts-ctrlo:
- Allow tests to run within container (debian:bookworm)
- Include scripts to use CtrlO repos and install dependencies as required
- WebDriver tests are currently disabled until it is ascertained how to run these within Debian container


droberts-ctrlo (commented):
Complete and ready for merge

PR #564, opened 2025-08-04
Formatted test and config code and updated test files

droberts-ctrlo:
Also removed extra yarn commands that weren't needed, and automated download of browserslist updates on build.

PR #565, opened 2025-08-04
Js formatting

droberts-ctrlo:
Also added ESLint to enforce code style for JS files


droberts-ctrlo (commented):
This currently requires merge of #564 before full review is completed - however, files have been reviewed where obvious changes are apparent

PR #566, opened 2025-08-04
Added JSDoc comments for development ease

droberts-ctrlo:
Also updated ESLint to ensure these are properly added and maintained


droberts-ctrlo:
Relies on #565

PR #572, opened 2025-08-07
Updated `match` endpoint to post and updated typeaheads and other code as appropriate

droberts-ctrlo:


droberts-ctrlo:
Filter typeaheads not currently working - will amend and submit commit.

droberts-ctrlo:
Filters fixed - all expected items tested and found to be working

PR #574, opened 2025-08-07
Workflow to build tests (and only test javascript files when they have changed)

droberts-ctrlo:
Relies on #563


droberts-ctrlo:
Closing in favour of a possible MakeFile.PL solution. Further investigation required.

PR #577, opened 2025-08-21
Upgraded React to v19

droberts-ctrlo:


droberts-ctrlo (commented):
Relies on #569 

Further review required once this is merged

PR #580, opened 2025-09-26
Complete default report setup (D 800)

droberts-ctrlo:
Create default report for download PDF link using GADS::PDFGenerator custom class


lib/GADS.pm
```diff
@@ -1948,8 +1948,10 @@ any qr{/(record|history|purge|purgehistory)/([0-9]+)} => require_login sub {
 
     if (defined param('pdf') && !$record->layout->no_download_pdf)
     {
-        my $pdf = $record->pdf->content;
-        return send_file(\$pdf, content_type => 'application/pdf', filename => "Record-".$record->current_id.".pdf" );
+        my $site = var 'site'
+            or error __"No site configured";
+        my $pdf = $record->pdf($site)->content;
+        return send_file( \$pdf, content_type => 'application/pdf', );
```
droberts-ctrlo: File name missing

lib/GADS/Record.pm
```diff
@@ -2721,80 +2721,126 @@ sub for_code
...
+
+    my $hdr_props = {
+        repeat    => 0,
+        justify   => 'center',
+        font_size => 12,
+        bg_color  => '#007c88',
+        fg_color  => '#ffffff',
+    };
+
+    my $result  = [$self->layout->all_user_read];
```
droberts-ctrlo: Copy/paste - why?

abeverley:
Thanks, looking good. Final few minor comments:
- Topics don't seem to necessarily be the correct order (I'm not sure why at this stage, just the test I have run)
- Can we line-up fat commas please?
- I think the PDF heading should be the same as it used to be ("Record 123") rather than the filename "Record-123.pdf"

droberts-ctrlo:
Fixed as per review - topics appear in the order they are in the layout view.

abeverley:
Thanks Dave, but I think the heading change is missing now? All good once that is updated.

PR #584, opened 2025-10-17
Updated to default to "all" view on view all button click (D795)

droberts-ctrlo:


abeverley:
Thanks Dave, looks good. My only comment is the heading for these records ("Linked Child Records"). The reason is that both "linked" and "child" records have other meanings in LinkSpace. Can we be more explicit and say something like "Records from field xxx of record nnn" or something like that?

abeverley:
Merged manually with commit f0423793

PR #585, opened 2025-10-22
Re-enable Add button on submit error [B: 1805]

Oliver-ctrlo:
Added a public setter to enable or disable the submit button.
Then included this in the modal submit error to unset the Disabled status of the button.


src/frontend/components/modal/modals/curval/lib/component.js
```diff
@@ -6,6 +6,7 @@ import { initializeRegisteredComponents } from 'component'
 import { validateRadioGroup, validateCheckboxGroup } from 'validation'
 import { fromJson } from 'util/common'
 import StorageProvider from 'util/storageProvider'
+import SubmitRecordButton from '../../../../../components/button/lib/submit-record-button.ts'
```
droberts-ctrlo, line 9: should be `import SubmitRecordButton from 'components/button/lib/submit-record-button.ts` - don't need the leading `..` at all.

PR #587, opened 2025-10-31
Paged chronology bs5

droberts-ctrlo:
Relies on #583 due to styling differences with BS4 version


droberts-ctrlo:
Closing in favour of a "better" solution

PR #589, opened 2025-11-21
Completed error logging to Audit (D990)

droberts-ctrlo:


lib/GADS/Audit.pm
```diff
@@ -70,7 +70,26 @@ has filtering => (
     builder => sub { +{} },
 );
 
-sub audit_types{ [qw/user_action login_change login_success logout login_failure/] };
+sub audit_types{ [qw/user_action login_change login_success logout login_failure script_error/] };
```
Oliver-ctrlo: Can we have new lines for each here?

src/frontend/js/lib/logging.js
```diff
@@ -1,34 +1,62 @@
...
+  formatMessage(type, ...message) {
+    let output = type + ': ';
+    for (let i = 0; i < message.length; i++) {
+      if (typeof message[i] === 'object') {
+        output += JSON.stringify(message[i]);
+      } else {
+        // This is wrapped so that anything that's not an object is converted to a string
+        output += `${message[i]}`;
       }
+      if (i < message.length - 1) output += ' ';
```
Oliver-ctrlo, line 57: Why do we need the extra space here at the end of the message?

src/frontend/js/lib/util/scriptErrorHandler/index.ts
```diff
@@ -0,0 +1,35 @@
+import { logging } from "logging";
+import { uploadMessage } from "./lib/MessageUploader";
+
+const createErrorString = (message: string, source: any, lineno: number, colno: number, error: Error | string | null) => {
```
Oliver-ctrlo, line 3: Can we have new lines here as-well?

src/frontend/js/lib/logging.js
Oliver-ctrlo: Consider using `message.join`.

src/frontend/js/lib/util/scriptErrorHandler/index.ts
droberts-ctrlo: No - convention in JS is single-line for methods (this is a constant method).

lib/GADS/API.pm
```diff
@@ -1464,6 +1464,15 @@ get '/api/get_key' => require_login sub {
     }
 };
 
+post '/api/script_error' => require_login sub {
+    my $body = _decode_json_body();
+
+    info "SCRIPT ERROR: " . $body->{url} . " - " . $body->{description};
```
abeverley: Might as well use the available interpolation:
```
info __x"SCRIPT ERROR: {url} - {description}",
    url => $body->{url}, description => $body->{description};
```

lib/GADS/API.pm
```diff
@@ -1464,6 +1464,15 @@ get '/api/get_key' => require_login sub {
...
+
+    content_type 'application/json; charset=UTF-8';
+    return success "Script error logged successfully";
```
abeverley: If you haven't already, can you test that this does what you think - I am not sure what will be returned from the endpoint with this code.

droberts-ctrlo: I don't quite know why I put `success` there - I've removed it now.

lib/GADS/API.pm
```diff
@@ -1464,6 +1464,16 @@ get '/api/get_key' => require_login sub {
...
 };
 
+post '/api/script_error' => require_login sub {
+    my $body = _decode_json_body();
+
+    info __x "SCRIPT ERROR: {url} - {description}",
+        url => $body->{url}, description => $body->{description};
+
+    content_type 'application/json; charset=UTF-8';
+    return "Script error logged successfully";
```
abeverley: Sorry, one final thing - you are setting the content-type to be JSON but not returning JSON. It doesn't matter which way around you fix it, you just need to ensure the content-type header and the actual content type match.

droberts-ctrlo: And *that's* why I used `success` - it was a typo, it was supposed to be `_success`!

PR #592, opened 2025-12-03
Fix views made with whitespace-only names (D1060)

pawelwlod:
A regex statement checks if the view name consists of just whitespace characters.
If true, it throws an error to the user.


abeverley:
Thanks Pawel, this looks great. I think it would just be worth adjusting the error message slightly, maybe to say "View name must not contain **only** whitespace characters"?

PR #594, opened 2025-12-17
Added appropriate error for blank field name heading in record imports (D1099)

pawelwlod:
Throw an error to the user if a field name heading is blank, alongside the column number in the CSV file being imported.

### DIFF
--- a/lib/GADS/Import.pm
+++ b/lib/GADS/Import.pm
@@ -301,7 +301,7 @@ sub _build_fields
             $search->{name_short} = $field if $self->short_names;
             my $f_rs = $self->schema->resultset('Layout')->search($search);
             error __x"Empty field name in import headings (Column {column})", column => $col_count
-                if !$field;
+                unless $field =~ /^\s*$/;
             error __x"Layout has more than one field named {name}", name => $field
                 if $f_rs->count > 1;
             error __x"Field '{name}' in import headings not found in table", name => $field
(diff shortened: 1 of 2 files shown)


lib/GADS/Import.pm
```diff
@@ -298,6 +300,8 @@ sub _build_fields
             $search->{name} = $field if !$self->short_names;
             $search->{name_short} = $field if $self->short_names;
             my $f_rs = $self->schema->resultset('Layout')->search($search);
+            error __x"Empty field name in import headings (Column {column})", column => $col_count
+                if !$field;
```
droberts-ctrlo: I would use `unless $field` rather than `if !$field` - this is more personal preference, so I would run with that

In mind of OT comment "If it's 0, what happens" it's possible to use `unless defined $field`

pawelwlod: Took a step further and accounted for all white-space characters using a regex statement.

abeverley:
Thanks Pawel, this is good. I realise the ticket says that this will be a manual test, but it would be relatively easy to add an automated test. Maybe you can take a look at that? A new section in `t/014_import.t`, creating a new `GADS::Import` object with appropriate CSV content and wrapping in a `try {}` block should suffice.

pwlodarski-ctrlo:
> Thanks Pawel, this is good. I realise the ticket says that this will be a manual test, but it would be relatively easy to add an automated test. Maybe you can take a look at that? A new section in `t/014_import.t`, creating a new `GADS::Import` object with appropriate CSV content and wrapping in a `try {}` block should suffice.

This has been actioned and committed to this PR.

PR #596, opened 2026-01-12
Update cache to ignore values over 250 characters (D1118)

droberts-ctrlo:
This code checks the `%values` hash and does not insert the value into the cache should there be any values over 250 characters inserted into the field.

### DIFF
--- a/lib/GADS/Datum/Code.pm
+++ b/lib/GADS/Datum/Code.pm
@@ -162,6 +162,8 @@ sub write_cache
     my $formatter = $self->schema->storage->datetime_parser;
 
     my @values = sort @{$self->value} if defined $self->value->[0];
+    # If the length is greater than 250 characters do not insert it to the cache table
+    return if grep { $_ > 250 } map { length($_) } @values;
 
     # We are generally already in a transaction at this point, but
     # start another one just in case
--- a/lib/GADS/Datum/Code.pm
+++ b/lib/GADS/Datum/Code.pm
@@ -122,6 +122,8 @@ sub _write_unique
     my $schema = $self->schema;
     if (my $table = $self->column->table_unique)
     {
+        # Return and don't write the cache if the value is over 250 characters
+        return if (grep { length($_) > 250 } values %values);
         $schema->storage->svp_begin("sp_uq_calc");
         try {
             $schema->resultset($table)->create({
@@ -162,8 +164,6 @@ sub write_cache
     my $formatter = $self->schema->storage->datetime_parser;
 
     my @values = sort @{$self->value} if defined $self->value->[0];
-    # If the length is greater than 250 characters do not insert it to the cache table
-    return if grep { $_ > 250 } map { length($_) } @values;
 
     # We are generally already in a transaction at this point, but
     # start another one just in case
--- a/t/007_code_cache.t
+++ b/t/007_code_cache.t
@@ -195,17 +195,28 @@ is_deeply([map $_->value_int, $urs->all], $expected, "Correct values for cached 
 
     $record->find_current_id(2);
 
+    $urs = $schema->resultset('CalcUnique')->search({ layout_id => $calc1->id });
+
+    $record->fields->{$string1->id}->set_value("Sandwiches");
+    $record->write(no_alerts => 1);
+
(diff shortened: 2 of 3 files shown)


pawelwlod:
> This code checks the `%values` hash and does not update insert the value into the cache should there be any values over 250 characters inserted into the field.

What do you mean by "update insert the value"?

droberts-ctrlo:
> > This code checks the `%values` hash and does not update insert the value into the cache should there be any values over 250 characters inserted into the field.
> 
> What do you mean by "update insert the value"?

Fixed

t/007_code_cache.t
```diff
@@ -11,6 +11,8 @@ use GADS::Schema;
 use lib 't/lib';
 use Test::GADS::DataSheet;
 
+use Data::Dump qw(pp);
```
pawelwlod: Is this a debugging dependency?

droberts-ctrlo: It is - I shall remove!

lib/GADS/Datum/Code.pm
```diff
@@ -162,6 +162,8 @@ sub write_cache
     my $formatter = $self->schema->storage->datetime_parser;
 
     my @values = sort @{$self->value} if defined $self->value->[0];
+    # If the length is greater than 250 characters do not insert it to the cache table
+    return \@values if grep { $_ > 250 } map { length($_) } @values;
```
abeverley: Unfortunately returning here means that the calculated value is not written to the database field at all. This means that any searches and rendering with prefetching will not work.

t/007_code_cache.t
```diff
@@ -181,4 +181,24 @@ is_deeply([map $_->value_int, $urs->all], $expected, "Correct values for cached
...
+    my $expected_calc_val = "This is a really long string, longer than before, to test caching of long strings should fail gracefully and not put any extra values in the cache, this is to test the edge case handling of long strings in cached unique values because they break the cache in some really weird ways sometimes";
+    $record->clear;
+    $record->find_current_id(1);
+    $calc1->code("
+        function evaluate (L1string1)
+            return L1string1
+        end
+    ");
+    $calc1->return_type('string');
+    $calc1->write;
```
abeverley, line 197: Might need to clear the layout here for the new field (see other examples) although no problem if it's working.

t/007_code_cache.t
```diff
@@ -181,4 +181,24 @@ is_deeply([map $_->value_int, $urs->all], $expected, "Correct values for cached
...
+    $record->fields->{$string1->id}->set_value($expected_calc_val);
+    $record->write(no_alerts => 1);
+
+    is_deeply([map $_->value_text, $urs->all], [qw/Bar7 Bar8/], "Correct values for cached table after long string write");
```
abeverley: Whilst this specifically is correct, we should also test whether the (now unused) calc unique fields are deleted.

We should also test whether the value in the `Calcval` table (probably better tested in `t/007_code.t`)

t/007_code_cache.t
```diff
@@ -181,4 +181,24 @@ is_deeply([map $_->value_int, $urs->all], $expected, "Correct values for cached
...
+
+    my $calc_val = $record->fields->{$string1->id}->as_string;
```
abeverley: This should be the calc field not the string field.

t/007_code_cache.t
```diff
@@ -181,4 +181,24 @@ is_deeply([map $_->value_int, $urs->all], $expected, "Correct values for cached
...
+    is($calc_val, $expected_calc_val, "Correct value returned from calc after long string write");
```
abeverley: I think we also need to add a test for `GADS::Records` searching on the long calc value (probably better placed in `t/007_code.t`. I think that will fail with the code as is (for the reasons above).

abeverley (commented):
Thanks Dave, good work on getting the test added. A bit more to do though I'm afraid...

abeverley:
Thanks Dave, this is great. Just 2 very minor things before merging: add a brief comment to the new test explaining what it is testing (I know I haven't always done this, and I've seen the problems it creates!); plus, remove the Data::Dumper from the included libraries.

PR #597, opened 2026-01-12
Add border and spacing to clarify where different text values end (D1135)

droberts-ctrlo:


pawelwlod:
Potentially add a description to e.g. explain which specific text values are affected.

abeverley:
Thanks Dave, looks good, but I am wondering whether this rule should be nested inside another one (that would be expected at that place on the page), just in case we start using `HTML::FromText` anywhere else in the app and don't want the separators?

droberts-ctrlo:
Because of the `not` rule, this won't display where we have a single value, but will where multiple values are used with that class - this will ensure that any section where there are multiple value fields that use this (not just text) the seperator is used and allows for clear demarcation.

PR #609, opened 2026-03-31
Changes the user view restrictions to an admin only view list (D:888)

Oliver-ctrlo:
Stops any non admin shared views from being displayed in the list of available options when applying the user view limits.


pwlodarski-ctrlo (commented):
Reviewed - all good!

droberts-ctrlo (commented):
Reviewed and all appears fine

PR #610, opened 2026-03-31
Fix record download when approval records exist (D:1350)

abeverley:


pwlodarski-ctrlo (commented):
PR reviewed - all is well.

PR #613, opened 2026-04-29
Remove logging line that was causing an error - HOTFIX - D1511

droberts-ctrlo:


droberts-ctrlo:
This was based off the incorrect branch - will re-do

PR #614, opened 2026-04-29
Remove logging line which was causing an error - HOTFIX - D1511

droberts-ctrlo:


pwlodarski-ctrlo (commented):
Hotfix PR is all good and ready for deployment.

PR #618, opened 2026-05-12
Logging fix - D1512

droberts-ctrlo:
fix error in logging.js where the message variable is re-declared in the else block of the logging code, causing the code to error when calling outside of a debug environment rather than undergoing the expected behaviour.

NB Unit test was removed due to adding excess complexity to the code - will re-evaluate and look at possible fix


pwlodarski-ctrlo (commented):
PR reviewed, all good.

droberts-ctrlo:
This fix was also implemented in the Bootstrap branch so is no longer needed

PR #620, opened 2026-05-14
Js refactor

droberts-ctrlo:


src/frontend/components/button/lib/submit-field-button.ts
```diff
@@ -1,5 +1,5 @@
 import 'jstree';
-import 'datatables.net';
+import 'datatables.net-bs4';
```
pwlodarski-ctrlo, line 2: Just to confirm, is this meant to be BS4 or BS5?

pwlodarski-ctrlo (commented):
PR Reviewed, apart from a singular clarification, all good!

src/frontend/components/button/lib/submit-field-button.ts
droberts-ctrlo: BS4 - BS5 will be in a few PR's time

PR #626, opened 2026-06-07
Fix alerts if user has view limit (#1531)

abeverley:


lib/GADS/Records.pm
```diff
@@ -634,79 +634,70 @@ sub search_view
...
-        view_id => $view->id
-    }, {
-        join => 'alerts',
-    })->all : (undef);
-
-    foreach my $user (@users)
+
+    my $filter  = $view->filter;
+    my $view_id = $view->id;
+    trace qq(About to decode filter for view ID $view_id);
```
pwlodarski-ctrlo, line 640: Debug statement?

droberts-ctrlo (commented):
Reviewed - no apparent issues

pwlodarski-ctrlo (commented):
PR reviewed - except a minor comment, all is well.

lib/GADS/Records.pm
abeverley: Thanks Pawel and good spot but this is actually a "proper" way of doing debugging in the code - these `trace` statements will only print anything when the log level is high enough.

pwlodarski-ctrlo: Understood, thanks for clarifying Andy!

droberts-ctrlo (commented):
Reviewed - all appears fine

pwlodarski-ctrlo (commented):
PR re-reviewed - all appears well.

PR #630, opened 2026-06-10
Fix where rebase was incorrect and missed files

droberts-ctrlo:
Bring dev-bootstrap up to date with:

- `datatables.net` changes which were missed in templates
- package upgrades to current versions
- JS fixes for package upgrades


pwlodarski-ctrlo (commented):
PR reviewed - all is well.

Noted that some tests are skipped due to an issue with Jest importing a certain library.

PR #631, opened 2026-06-10
React upgrades

droberts-ctrlo:
Reliant on #630


droberts-ctrlo:
Closing as it hasn't pulled throught the correct change set.

PR #634, opened 2026-06-15
React upgrades

droberts-ctrlo:


src/frontend/components/dashboard/lib/react/Dashboard/DashboardView.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import RGL, { WidthProvider } from 'react-grid-layout'; // Do not go over v1 for now
```
pwlodarski-ctrlo, line 2: Is there a comparability issue?

src/frontend/components/dashboard/lib/react/EditModal/EditModal.tsx
```diff
@@ -0,0 +1,69 @@
+/* eslint-disable @stylistic/semi */
+import React, { useEffect } from 'react';
+import Modal, {Styles} from 'react-modal';
+import { AppModalProps } from '../types';
+
+/**
+ * Edit modal component
+ * @param {AppModalProps} props - The modal props
+ * @returns {React.JSX.Element} The rendered modal component
+ * @todo: A lot of state here - I think I will revisit this later
```
pwlodarski-ctrlo: A lot of state?

pwlodarski-ctrlo (commented):
PR reviewed - two clarification comments added.
Apart from that, all seems well.

src/frontend/components/dashboard/lib/react/Dashboard/DashboardView.tsx
droberts-ctrlo: I assume you mean compatibility?

Yes, there is. The changes between v1 and v2 are "breaking" changes - this means that we would have to re-write a considerable amount of the dashboard code. The version (v2) came out after the upgrades were completed (the first time) and it is felt that rather than have to re-write a large chunk of the codebase, it's best to leave it as-is for now.

src/frontend/components/dashboard/lib/react/EditModal/EditModal.tsx
droberts-ctrlo: This comment shouldn't have been left in - removed now

PR #636, opened 2026-06-19
Fix graphs not allowing grouping with no date (D1682)

droberts-ctrlo:


pwlodarski-ctrlo (commented):
PR reviewed - all is well.

droberts-ctrlo:
Closing as not correct fix - problem is more complex than originally assumed

PR #638, opened 2026-06-22
Add wording for file uploads and update templates (D1092)

droberts-ctrlo:


pwlodarski-ctrlo (commented):
PR reviewed - all is well.

PR #641, opened 2026-07-16
Script to import and export layout group permissions in bulk [D: 1825]

Oliver-ctrlo:
This script allows bulk importing and exporting group layout permissions.

### DIFF
--- a/bin/bulk_layout_permissions.pl
+++ b/bin/bulk_layout_permissions.pl
@@ -1,0 +1,121 @@
+#!/usr/bin/perl
+
+use FindBin;
+use lib "$FindBin::Bin/../lib";
+
+use Dancer2;
+use Dancer2::Plugin::DBIC;
+use Getopt::Long;
+use Text::CSV;
+use feature 'say';
+
+my ($action);
+
+GetOptions(
+    'action=s'   => \$action
+) or exit;
+
+$action or die "ERROR: Please state if you want to import/export group permissions with --action";
+
+if ($action eq 'import') {
+    Import_permissions();
+}
+elsif ($action eq 'export') {
+    Export_permissions();
+}
+else {
+    die "ERROR: You must either import or export within --action";
+}
+
+sub Import_permissions {
+    say "Enter the file you want to import from:";
+    my $file = <STDIN>;
+    chomp($file);
+
+    -f $file or die "ERROR: File '$file' does not exist";
+
+    $file or die "Usage: $0 filename";
+    my $csv = Text::CSV->new({ binary => 1 })
+    or die "Cannot use CSV: ".Text::CSV->error_diag ();
+    open my $fh, "<:encoding(utf8)", $file or die "$file: $!";
+    my $guard = schema->txn_scope_guard;
+     
+    while (my $row = $csv->getline($fh)) {
+        my ($layout_id, $group_id, $permission) = @$row;
+        my $RowCount = $csv->record_number;
+        
+        unless (defined $layout_id && $layout_id =~ /^\d+$/) {
+            die "ERROR: Invalid Layout ID '$layout_id' on row $RowCount.\n";
+        }
+        unless (defined $group_id && $group_id =~ /^\d+$/) {
+            die "ERROR: Invalid Group ID '$group_id' on row $RowCount.\n";
+        }
+        unless (defined $permission && $permission =~ /^(?:read|write_new|write_existing|write_existing_no_approval|write_new_no_approval)$/) {
+            die "ERROR: Invalid Permission '$permission' on row $RowCount. Value must be read, write_new, write_existing, or write_existing_no_approval.\n";
+        }
+            
+        rset('Layout')->find($layout_id) or die "ERROR: Layout ID '$layout_id' does not exist\n";
+        rset('Group')->find($group_id)   or die "ERROR: Group ID '$group_id' does not exist\n";
+            
+        my $existing = rset('LayoutGroup')->find({
+            layout_id  => $layout_id,
+            group_id   => $group_id,
+            permission => $permission
+        });
+        
+        if ($existing) {
+            say "Skipping permission '$permission' for group '$group_id' on layout '$layout_id' because this permission already exists";
+            next;
+        }
+        
+        rset('LayoutGroup')->create({
+		        layout_id  => $layout_id,
+		        group_id   => $group_id,
+		        permission => $permission,
+        });
+        say "Will add permission '$permission' to group '$group_id' for layout '$layout_id'";
+    }
+    $guard->commit;
+    say "Import complete";
+}
+
+sub Export_permissions {
+    my $export_file = './Group_permission_export.csv';
+    ! -f $export_file or die "ERROR: $export_file already exists, unable to overwrite file!";
+    say "Enter the group IDs you're looking to export (e.g. 1,2,3): ";
+    my $input = <STDIN>;
+    chomp($input);
+
+    if (!$input || $input !~ /^\s*\d+(?:\s*,\s*\d+)*\s*$/) {
+        die "Error: You need to enter a number or a comma-separated list of numbers.\n";
+    }
+ 
+    my @groups = split(/\s*,\s*/, $input);
+
+    my $csv = Text::CSV->new({ binary => 1, auto_diag => 1 });
+    open my $fh, ">:encoding(utf8)", $export_file or die "$export_file: $!";
+    
+    my @layout_groups = rset('LayoutGroup')->search(
+        {
+            'me.group_id' => { '-in' => \@groups },
+        },
+        {
+            prefetch => [ 'layout' ],
+        }
+    )->all;
+    
+    $csv->say($fh, [ 'Instance ID','Layout Name','Layout ID', 'Group ID','Permission' ]);
+    
+    foreach my $lg (@layout_groups) {
+        my $layout      = $lg->layout;
+        my $instance_id = $layout->instance_id;
+        my $layout_name = $layout->name ;
+        my $group_id    = $lg->group_id;
+        my $layout_id   = $lg->layout_id;
+        my $permission  = $lg->permission;
+
+        $csv->say($fh, [ $instance_id, $layout_name, $layout_id, $group_id, $permission ]);
+    }
+    close($fh);
+    say "Successfully exported permissions to '$export_file'";
+}
--- a/bin/bulk_layout_permissions.pl
+++ b/bin/bulk_layout_permissions.pl
@@ -13,7 +13,7 @@ use Log::Report syntax => 'LONG';
 my ($action, $file);
 
 GetOptions(
-    'action=s'   => \$action,
+    'action=s' => \$action,
     'file=s'   => \$file
 ) or exit;
 
@@ -32,11 +32,11 @@ else {
 
 sub Import_permissions {
 
-    -f $file or report ERROR => "File '$file' does not exist";
+    -f $file or report ERROR => "File '{file}' does not exist", file => $file;
 
     my $csv = Text::CSV->new({ binary => 1 })
-    or report ERROR => "Cannot use CSV: ".Text::CSV->error_diag ();
-    open my $fh, "<:encoding(utf8)", $file or report FAULT => "$file";
+        or report ERROR => "Cannot use CSV: ".Text::CSV->error_diag ();
+    open my $fh, "<:encoding(utf8)", $file or report FAULT => "Unable to open file {file}", file => $file;
     my $guard = schema->txn_scope_guard;
     
     my $headers = $csv->getline($fh);
@@ -47,7 +47,7 @@ sub Import_permissions {
 
     while (my $row = $csv->getline_hr($fh)) {
 	    
-        my $RowCount = $csv->record_number;
+    my $RowCount = $csv->record_number;
 	my $layout_id = $row->{layout_id}; 
 	my $group_id = $row->{group_id};
 	my $permission = $row->{permission};
--- a/bin/bulk_layout_permissions.pl
+++ b/bin/bulk_layout_permissions.pl
@@ -8,15 +8,17 @@ use Dancer2::Plugin::DBIC;
 use Getopt::Long;
 use Text::CSV;
 use feature 'say';
+use Log::Report syntax => 'LONG';
 
-my ($action);
+my ($action, $file);
 
 GetOptions(
-    'action=s'   => \$action
+    'action=s'   => \$action,
+    'file=s'   => \$file
 ) or exit;
 
-$action or die "ERROR: Please state if you want to import/export group permissions with --action";
-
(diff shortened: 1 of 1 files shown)


bin/bulk_layout_permissions.pl
```diff
@@ -0,0 +1,121 @@
...
+        my $RowCount = $csv->record_number;
+        
+        unless (defined $layout_id && $layout_id =~ /^\d+$/) {
+            die "ERROR: Invalid Layout ID '$layout_id' on row $RowCount.\n";
+        }
+        unless (defined $group_id && $group_id =~ /^\d+$/) {
+            die "ERROR: Invalid Group ID '$group_id' on row $RowCount.\n";
+        }
+        unless (defined $permission && $permission =~ /^(?:read|write_new|write_existing|write_existing_no_approval|write_new_no_approval)$/) {
+            die "ERROR: Invalid Permission '$permission' on row $RowCount. Value must be read, write_new, write_existing, or write_existing_no_approval.\n";
```
pwlodarski-ctrlo: Would `write_new_no_approval` be included too?

bin/bulk_layout_permissions.pl
```diff
@@ -0,0 +1,121 @@
...
+elsif ($action eq 'export') {
+    Export_permissions();
+}
+else {
+    die "ERROR: You must either import or export within --action";
+}
+
+sub Import_permissions {
+    say "Enter the file you want to import from:";
+    my $file = <STDIN>;
```
droberts-ctrlo: Could this not be a script argument?

bin/bulk_layout_permissions.pl
```diff
@@ -0,0 +1,121 @@
...
+		        permission => $permission,
+        });
+        say "Will add permission '$permission' to group '$group_id' for layout '$layout_id'";
+    }
+    $guard->commit;
+    say "Import complete";
+}
+
+sub Export_permissions {
+    my $export_file = './Group_permission_export.csv';
```
pwlodarski-ctrlo: Would it be beneficial for the filename to be unique? E.g. it could have the group IDs or the current datetime appended.

bin/bulk_layout_permissions.pl
```diff
@@ -0,0 +1,121 @@
...
+    -f $file or die "ERROR: File '$file' does not exist";
+
+    $file or die "Usage: $0 filename";
+    my $csv = Text::CSV->new({ binary => 1 })
+    or die "Cannot use CSV: ".Text::CSV->error_diag ();
+    open my $fh, "<:encoding(utf8)", $file or die "$file: $!";
+    my $guard = schema->txn_scope_guard;
+     
+    while (my $row = $csv->getline($fh)) {
+        my ($layout_id, $group_id, $permission) = @$row;
```
droberts-ctrlo: This assumes CSV is set up correctly - you can import the rows as a hash containing `{column1 => value, column2 => value ... }` where column1, column2, etc are the header of the file which would mitigate if the file data is in the incorrect order (currently, it will assume it's correct and at worst, import the data incorrectly, at best, die)

bin/bulk_layout_permissions.pl
```diff
@@ -0,0 +1,121 @@
...
+}
+elsif ($action eq 'export') {
+    Export_permissions();
+}
+else {
+    die "ERROR: You must either import or export within --action";
+}
+
+sub Import_permissions {
+    say "Enter the file you want to import from:";
```
droberts-ctrlo: Maybe set up the option to allow for logging here as well? Output information to a log file to show what's been done? Otherwise there's no audit trail and if something goes wrong we won't easily know what's borken.

bin/bulk_layout_permissions.pl
```diff
@@ -0,0 +1,121 @@
...
+        rset('Group')->find($group_id)   or die "ERROR: Group ID '$group_id' does not exist\n";
+            
+        my $existing = rset('LayoutGroup')->find({
+            layout_id  => $layout_id,
+            group_id   => $group_id,
+            permission => $permission
+        });
+        
+        if ($existing) {
+            say "Skipping permission '$permission' for group '$group_id' on layout '$layout_id' because this permission already exists";
```
droberts-ctrlo: Do we need to say anything here? Could just put `next if $existing`

bin/bulk_layout_permissions.pl
```diff
@@ -0,0 +1,121 @@
...
+else {
+    die "ERROR: You must either import or export within --action";
+}
+
+sub Import_permissions {
+    say "Enter the file you want to import from:";
+    my $file = <STDIN>;
+    chomp($file);
+
+    -f $file or die "ERROR: File '$file' does not exist";
```
droberts-ctrlo: Should use `error` rather than `die` here (remember to import logreport)

bin/bulk_layout_permissions.pl
```diff
@@ -0,0 +1,121 @@
...
+
+    $file or die "Usage: $0 filename";
+    my $csv = Text::CSV->new({ binary => 1 })
+    or die "Cannot use CSV: ".Text::CSV->error_diag ();
```
droberts-ctrlo: Should use `error` rather than `die` here

bin/bulk_layout_permissions.pl
```diff
@@ -0,0 +1,121 @@
...
+    open my $fh, ">:encoding(utf8)", $export_file or die "$export_file: $!";
+    
+    my @layout_groups = rset('LayoutGroup')->search(
+        {
+            'me.group_id' => { '-in' => \@groups },
+        },
+        {
+            prefetch => [ 'layout' ],
+        }
+    )->all;
```
pwlodarski-ctrlo, line 112: Would it be worth moving the `$guard->commit;` to below this search, just in case any exceptions are thrown?

bin/bulk_layout_permissions.pl
```diff
@@ -0,0 +1,121 @@
...
+    or die "Cannot use CSV: ".Text::CSV->error_diag ();
+    open my $fh, "<:encoding(utf8)", $file or die "$file: $!";
+    my $guard = schema->txn_scope_guard;
+     
+    while (my $row = $csv->getline($fh)) {
+        my ($layout_id, $group_id, $permission) = @$row;
+        my $RowCount = $csv->record_number;
+        
+        unless (defined $layout_id && $layout_id =~ /^\d+$/) {
+            die "ERROR: Invalid Layout ID '$layout_id' on row $RowCount.\n";
```
droberts-ctrlo: Should use `error` here - also can be `(defined $layout_id && $layout_id =~ /^\d+$/) || error ...` - the second point is more of a symantic difference and down to code-style and preference.

pwlodarski-ctrlo (commented):
PR reviewed - couple comments added for consideration.

bin/bulk_layout_permissions.pl
```diff
@@ -0,0 +1,121 @@
...
+        }
+        unless (defined $group_id && $group_id =~ /^\d+$/) {
```
droberts-ctrlo, line 58: As above

bin/bulk_layout_permissions.pl
```diff
@@ -0,0 +1,121 @@
...
+            die "ERROR: Invalid Group ID '$group_id' on row $RowCount.\n";
+        }
+        unless (defined $permission && $permission =~ /^(?:read|write_new|write_existing|write_existing_no_approval|write_new_no_approval)$/) {
```
droberts-ctrlo, line 61: As above

bin/bulk_layout_permissions.pl
```diff
@@ -0,0 +1,121 @@
...
+            die "ERROR: Invalid Permission '$permission' on row $RowCount. Value must be read, write_new, write_existing, or write_existing_no_approval.\n";
+        }
+            
+        rset('Layout')->find($layout_id) or die "ERROR: Layout ID '$layout_id' does not exist\n";
```
droberts-ctrlo: Should be `error`, not `die`

bin/bulk_layout_permissions.pl
droberts-ctrlo: Parameter? use `--file` or something and only set this if it's not set as a parameter (i.e. `$export_file //= './Group_permission_export.csv'`

bin/bulk_layout_permissions.pl
```diff
@@ -0,0 +1,121 @@
...
+        });
+        say "Will add permission '$permission' to group '$group_id' for layout '$layout_id'";
+    }
+    $guard->commit;
+    say "Import complete";
+}
+
+sub Export_permissions {
+    my $export_file = './Group_permission_export.csv';
+    ! -f $export_file or die "ERROR: $export_file already exists, unable to overwrite file!";
```
droberts-ctrlo: Should be `error` rather than `die`

bin/bulk_layout_permissions.pl
```diff
@@ -0,0 +1,121 @@
...
+    say "Enter the group IDs you're looking to export (e.g. 1,2,3): ";
+    my $input = <STDIN>;
+    chomp($input);
+
+    if (!$input || $input !~ /^\s*\d+(?:\s*,\s*\d+)*\s*$/) {
+        die "Error: You need to enter a number or a comma-separated list of numbers.\n";
```
droberts-ctrlo: Should be `error` rather than `die`

bin/bulk_layout_permissions.pl
```diff
@@ -0,0 +1,121 @@
...
+    }
+ 
+    my @groups = split(/\s*,\s*/, $input);
+
+    my $csv = Text::CSV->new({ binary => 1, auto_diag => 1 });
+    open my $fh, ">:encoding(utf8)", $export_file or die "$export_file: $!";
```
droberts-ctrlo: Should be `fault` rather than `die` - this will also remove the need for `$!`

droberts-ctrlo (commented):
Various observations - the majority of which are from the use of `die` rather than `error` or `fault` - you can also use `info` rather than `say`, but that's a matter of personal preference and not an issue.

bin/bulk_layout_permissions.pl
droberts-ctrlo: That permission is a left over from GADS 1.0 and is no longer used

bin/bulk_layout_permissions.pl
droberts-ctrlo: No - we want the guard to fall out of scope and therefore roll back any and all changes, rather than doing a partial commit if only some changes are made before the error.

bin/bulk_layout_permissions.pl
Oliver-ctrlo: Changed to a script argument instead so it would be the same for both import and export.

bin/bulk_layout_permissions.pl
Oliver-ctrlo: Agreed and done.

bin/bulk_layout_permissions.pl
Oliver-ctrlo: Would rather keep this so every row results in some kind of output even if nothing changes. Just makes it easier to track and follow progress.

bin/bulk_layout_permissions.pl
```diff
@@ -0,0 +1,128 @@
...
+use Dancer2::Plugin::DBIC;
+use Getopt::Long;
+use Text::CSV;
+use feature 'say';
+use Log::Report syntax => 'LONG';
+
+my ($action, $file);
+
+GetOptions(
+    'action=s'   => \$action,
```
pwlodarski-ctrlo, line 16: May be worth aligning the fat commas to keep the code neat.

bin/bulk_layout_permissions.pl
```diff
@@ -0,0 +1,128 @@
...
+    open my $fh, "<:encoding(utf8)", $file or report FAULT => "$file";
+    my $guard = schema->txn_scope_guard;
+    
+    my $headers = $csv->getline($fh);
+    $csv->column_names(@$headers) if $headers;
+    
+    report ERROR => "Invalid CSV header/s. Import must consist of 3 columns: layout_id, group_id and permission" 
+       unless $headers && join(',', sort @$headers) eq 'group_id,layout_id,permission';
+
+    while (my $row = $csv->getline_hr($fh)) {
```
pwlodarski-ctrlo, line 48: The indentation of some lines within this `while` loop should be reviewed.

bin/bulk_layout_permissions.pl
```diff
@@ -0,0 +1,128 @@
...
+use Log::Report syntax => 'LONG';
+
+my ($action, $file);
+
+GetOptions(
+    'action=s'   => \$action,
+    'file=s'   => \$file
+) or exit;
+
+$action or report ERROR => "Please state if you want to import/export group permissions with --action";
```
pwlodarski-ctrlo, line 20: May be worth aligning the fat commas for clarity across the script - it's done in other places in the stack too.

bin/bulk_layout_permissions.pl
```diff
@@ -0,0 +1,128 @@
...
+    report ERROR => "You must state if you want to import or export with --action";
+}
+
+
+sub Import_permissions {
+
+    -f $file or report ERROR => "File '$file' does not exist";
+
+    my $csv = Text::CSV->new({ binary => 1 })
+    or report ERROR => "Cannot use CSV: ".Text::CSV->error_diag ();
```
pwlodarski-ctrlo, line 38: For clarity, would be ideal to indent this line.

pwlodarski-ctrlo (commented):
PR re-reviewed - minor comments added.

bin/bulk_layout_permissions.pl
```diff
@@ -0,0 +1,128 @@
...
+    Export_permissions();
+}
+else {
+    report ERROR => "You must state if you want to import or export with --action";
+}
+
+
+sub Import_permissions {
+
+    -f $file or report ERROR => "File '$file' does not exist";
```
abeverley, line 35: Generally better to use parameters, e.g. `ERROR => "File '{file}' does not exist", file => $file`

bin/bulk_layout_permissions.pl
```diff
@@ -0,0 +1,128 @@
...
+
+    my $csv = Text::CSV->new({ binary => 1 })
+    or report ERROR => "Cannot use CSV: ".Text::CSV->error_diag ();
+    open my $fh, "<:encoding(utf8)", $file or report FAULT => "$file";
```
abeverley, line 39: Better to write something like `FAULT => "Unable to open file {file}", file => $file` - the reason will be automatically appended when using `FAULT`

abeverley:
Looks good to me to go ahead and use. Various comments to tidy it up in due course. Ideally these sorts of scripts should use high-level functions in the library files (so that as the internal API changes the scripts don't need to be updated). However, this would be a bigger piece of work, as the libraries would also need to be updated for the purpose of this script.

pwlodarski-ctrlo (commented):
PR re-reviewed - changes look fine to me.

PR #645, opened 2026-08-19
Change view type to a radio so that only one type is selectable (D888)

droberts-ctrlo:


src/frontend/components/form-group/radio-group/radio-reveal/lib/component.ts
```diff
@@ -0,0 +1,39 @@
...
+        this.el = $(this.element);
+        this.init();
+    }
+
+    private init() {
+        const target = this.el.data("radio-target");
+        const value = this.el.data("radio-value");
+
+        if (!target || !value) {
+            console.error("RadioRevealComponent: Missing data-radio-target or data-radio-value attribute.");
```
pwlodarski-ctrlo: Is this picked up by our `/scripterror` frontend error catching endpoint?

pwlodarski-ctrlo (commented):
PR reviewed - appears fine, just one minor clarification comment.

droberts-ctrlo:
On discussion with JH we will also try using a dropdown rather than radio buttons to see if this is a better design.

PR #647, opened 2026-08-21
Added created_by field to users (D1178)

pwlodarski-ctrlo:
When a user is created or an account request is approved, the creator/approver's username is set as the `created_by` value within the user's record. This is likewise reflected in the appropriate audit log messages.


bin/seed-database.pl
```diff
@@ -87,9 +87,11 @@
 
 say qq(Creating initial username "$initial_username"...);
 my $user = rset('User')->create({
-    username => $initial_username,
-    email    => $initial_username,
-    site_id  => $site->id,
+    username   => $initial_username,
+    email      => $initial_username,
+    site_id    => $site->id,
```
droberts-ctrlo, line 92: This should be set automatically - this is in `/lib/GADS/Schema.pm`

lib/GADS/Schema.pm
```diff
@@ -8,7 +8,7 @@ use base 'DBIx::Class::Schema';
 
 __PACKAGE__->load_namespaces;
 
-our $VERSION = 111;
+our $VERSION = 112;
```
droberts-ctrlo, line 11: This will need updating to 113 if it is merged after calc datetime (sorry)

droberts-ctrlo (commented):
Nothing major - just a couple of (very) minor comments

bin/seed-database.pl
pwlodarski-ctrlo: The `site_id` was already part of the script before these changes - only the indenting was updated on that line.

lib/GADS/Schema.pm
pwlodarski-ctrlo: Ouch. After that PR is merged, I'll merge the changes into this branch and upgrade the schema.

abeverley:
Thanks Pawel. I've manually merged this in the interest of managing schema changes. Some comments for follow up in slow time:

- I'm wondering what the reason is for falling back to "email" in the audit logs, I would have thought "username" is always defined?
- I've changed the created by column to a foreign key referencing the actual user for better data integrity
- As a result of the above change, I've removed the created by from the CSV download for ease, as it requires joining and fetching at the same time. This can be added later if need be
- Also as a result of the above, the ordering doesn't entirely work as expected (it sorts on id rather than display name)

